PR #1, opened 2019-12-13
Change the port mapping in docker compose to 8080

biranjan:
If you build the docker image from the docker file  HTTP port is exposed at 8080 however docker-compose maps the port to 5004.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     image: "metadata_service:latest"
     container_name: "metadata_service"
     ports:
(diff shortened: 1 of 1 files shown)


eliorc:
Indeed, the port exposed by the file is 8080, so if using the `docker-compose.yml` file as is, it fails to work. Was really difficult to understand why isn't it working, thanks to this pull request I understood this was a mistake.

ferras:
@biranjan thank you for your contribution
This pull request should address your concern. The port and the host is now configurable.

PR #4, opened 2020-01-28
make host and port for metadata service configurable

ferras:


ferras:
I will update this pull request when i deploy the latest image of the service to dockerhub. The image hosted on dockerhub would need to be updated to leverage the new env variables

ferras:
latest image has been pushed to dockerhub

PR #8, opened 2020-05-14
only return artifacts from latest attempt

ferras:
ensures that only the artifacts from the latest attempt are returned

see github issue:
https://github.com/Netflix/metaflow/issues/201


savingoyal:
@ferras are we good to release this?

PR #30, opened 2020-10-02
modify primary key

ferras:


msavela:
I believe the migration should be included here as well: https://github.com/Netflix/metaflow-service/blob/master/services/migration_service/api/__init__.py#L6

ferras:
> I believe the migration should be included here as well: https://github.com/Netflix/metaflow-service/blob/master/services/migration_service/api/__init__.py#L6

yep, diff isn't ready to be merged in yet. Still testing to determine if we should make this a major or minor version upgrade (although the change doesn't break the api, the migration may cause some downtime on the db since we are touching the primary key.

Dockerfile
```diff
@@ -7,9 +7,11 @@ RUN apt-get update && apt-get -y install python3.7 && apt-get -y install python3
 RUN pip3 install virtualenv && pip3 install requests
 
 RUN virtualenv /opt/v_1_0_1 -p python3
+RUN virtualenv /opt/v_2_0_4 -p python3
 RUN virtualenv /opt/latest -p python3
 
 RUN /opt/v_1_0_1/bin/pip install https://github.com/Netflix/metaflow-service/archive/1.0.1.zip
+RUN /opt/v_2_0_4/bin/pip install https://github.com/Netflix/metaflow-service/archive/2.0.4.zip
```
ferras, line 14: actually I will revert this change. we can just have two goose versions point to latest

msavela (approved):

services/migration_service/migration_files/20201002000616_update_metadata_primary_key.sql
```diff
@@ -0,0 +1,43 @@
+-- +goose Up
```
saikonen, line 1: can you also make sure that migrations apply successfully in development mode, where metadata service has possibly created tables before migration service runs.

ferras: changes in this migration file would still work even if the db was initially created by the metadata service

PR #31, opened 2020-10-07
Add task heartbeat monitor

saikonen:
- refactors heartbeat monitor into an abstract for shared functionality, also documenting the common interface
- adds Task heartbeat monitor. Accounts for attempt_id when expiring task heartbeats.
- add test coverage for tasks with heartbeat

### DIFF
--- a/services/ui_backend_service/api/heartbeat_monitor.py
+++ b/services/ui_backend_service/api/heartbeat_monitor.py
@@ -78,6 +78,8 @@ class RunHeartbeatMonitor(HeartbeatMonitor):
       self.remove_from_watch(run_id)
     
   async def add_to_watch(self, run_id):
+    # TODO: Optimize db trigger so we do not have to fetch a record in order to add it to the
+    # heartbeat monitor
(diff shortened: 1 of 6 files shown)


services/ui_backend_service/api/heartbeat_monitor.py
```diff
@@ -63,25 +99,88 @@ async def get_run(self, run_id):
...
+    
+  async def add_to_watch(self, data):
+    # TODO: Optimize db trigger so we do not have to fetch a record in order to add it to the
+    # heartbeat monitor
+    task = await self.get_task(
+      data["flow_id"],
+      data["run_number"],
+      data["step_name"],
+      data["task_id"]
+    )
```
saikonen, line 144: There is a performance concern related to the DB trigger_notify not carrying all necessary info in order to add a task to be monitored. Critically the `attempt_id` is missing from the trigger, so we need to perform a db fetch because of this.

msavela (approved):

msavela (approved):

PR #33, opened 2020-10-12
Fix task attempt timestamps

saikonen:
* adds logic to query for using the attempt and attempt-done metadata timestamps for task attempt started_at/finished_at
* also includes fallback logic to support old task data that do not have the correct metadata, due to incorrect constraints in older metadata services
* improvements to test helper output
* fixes broken task tests due to new model property.
* refactoring some of the task tests for better readability. adding new test cases for newer metadata tasks.

requires #30 to be merged before this for tests to pass.


msavela (approved):

services/data/postgres_async_db.py
```diff
@@ -668,33 +668,71 @@ class AsyncTaskTablePostgres(AsyncPostgresTable):
...
+                attempt.field_name = 'attempt' AND
+                task_ok.attempt_id = attempt.value::int
+            )
+            LEFT JOIN {metadata_table} as done ON (
+                task_ok.flow_id = done.flow_id AND
+                task_ok.step_name = done.step_name AND
+                task_ok.task_id = done.task_id AND
+                done.field_name = 'attempt-done' AND
+                task_ok.attempt_id = done.value::int
+            )
```
saikonen, line 696: if absolute accuracy isn't a concern in favor of performance, this join could be omitted, and we can simply use the task_ok.ts_epoch as the finished_at in every case.

PR #35, opened 2020-10-15
Add artifact refinement for database responses

saikonen:
* Add generic postprocessing to enrich database responses with content from S3.
* change find_records postprocessing helper to accept async callbacks
* adds `task_ok` s3 location to Task database responses, that gets postprocessed to the actual content.


services/data/postgres_async_db.py
```diff
@@ -171,7 +172,8 @@ async def get_records(self, filter_dict={}, fetch_single=False,
 
     async def find_records(self, conditions: List[str] = None, values=[], fetch_single=False,
                            limit: int = 0, offset: int = 0, order: List[str] = None, groups: List[str] = None,
-                           group_limit: int = 10, expanded=False, enable_joins=False) -> (DBResponse, DBPagination):
+                           group_limit: int = 10, expanded=False, enable_joins=False,
+                           postprocess: Callable[[DBResponse], DBResponse] = None) -> (DBResponse, DBPagination):
```
saikonen, line 176: @ferras any opinions for/against introducing the record postprocessing capability to the data module like this? UI backend uses postprocessing for artifacts, and with this PR also for task route responses (refining responses with S3 content).

msavela (approved):

PR #36, opened 2020-10-15
Websocket past data: `since` to return missing data

msavela:
Introduces ability to deliver past Websocket events while subscribing to resources (defined by `since` parameter):

```json
  {
    "type": "SUBSCRIBE", 
    "resource": "path-to-subscribable-restful-resource",
    "uuid": "client-generated-uuid",
    "since": 1602752197
  }
```

This event queue TTL is 5 minutes by default and can be controlled with `WS_QUEUE_TTL_SECONDS` environment variable. This TTL is used by the UI as well to determine if client data is stale in case reconnecting took too long (and prompts user to refresh browser if that's the case).

Added 3 integration tests:
- `ws_test.test_subscription_queue` - Make sure past data is returned correctly when `since` parameter is provided
- `ws_test.test_subscription_queue_without_since` - Make sure past data is not returned in case `since` parameter not provided
- `ws_test.test_subscription_queue_ttl_expired` - Make sure past data is not returned when queue is items are past TTL

This PR also includes `README.md` entry for `PREFETCH_RUNS_LIMIT` environment variable.

### DIFF
--- a/services/ui_backend_service/api/utils.py
+++ b/services/ui_backend_service/api/utils.py
@@ -303,11 +303,11 @@ class TTLQueue:
         self._ttl: int = ttl_in_seconds
         self._queue = deque()
 
-    def append(self, value: any):
+    async def append(self, value: any):
         self._queue.append((time.time(), value))
-        self.discard_expired_values()
+        await self.discard_expired_values()
 
-    def discard_expired_values(self):
+    async def discard_expired_values(self):
         cutoff_time = time.time() - self._ttl
         try:
             while self._queue[0][0] < cutoff_time:
@@ -315,9 +315,9 @@ class TTLQueue:
         except IndexError:
             pass
 
-    def values(self):
-        self.discard_expired_values()
+    async def values(self):
+        await self.discard_expired_values()
         return self._queue
 
-    def values_since(self, since_epoch: int):
-        return [value for value in self.values() if value[0] >= since_epoch]
+    async def values_since(self, since_epoch: int):
+        return [value for value in await self.values() if value[0] >= since_epoch]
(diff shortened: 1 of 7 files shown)


services/ui_backend_service/api/utils.py
```diff
@@ -294,3 +296,28 @@ async def find_records(request: web.BaseRequest, async_table=None, initial_condi
...
+        self._queue.append((time.time(), value))
+        self.discard_expired_values()
+
+    def discard_expired_values(self):
+        cutoff_time = time.time() - self._ttl
+        try:
+            while self._queue[0][0] < cutoff_time:
+                self._queue.popleft()
+        except IndexError:
+            pass
```
saikonen, line 316: clear choice to keep these synchronous instead of async? Is there any concern that the queue pop might be sluggish if it is packed with events during high utilization times?

services/ui_backend_service/api/ws.py
```diff
@@ -43,29 +48,47 @@ def __init__(self, app, event_emitter=None):
...
-                            data=data, conditions=sub.conditions, values=sub.values)
-                    else:
-                        filters_match_request = True
-                    if filters_match_request:
-                        payload = {'type': operation, 'uuid': sub.uuid,
-                                   'resource': resource, 'data': data}
-                        await sub.ws.send_str(json.dumps(payload))
-
-    async def subscribe_to(self, ws, uuid: str, resource: str):
+        # Append to QUEUE so that we can send events to users in case of
```
saikonen, line 51: comment incomplete

saikonen (approved):

saikonen (approved):

PR #50, opened 2020-11-12
Codestyle fixes

saikonen:
* fixes PEP8 code style errors in metadata_service, migration_service, ui_backend_service and services/utils
* disable continue-on-error for codestyle test

### DIFF
--- a/services/utils/__init__.py
+++ b/services/utils/__init__.py
@@ -15,6 +15,7 @@ version = pkg_resources.require("metadata_service")[0].version
 METADATA_SERVICE_VERSION = version
 METADATA_SERVICE_HEADER = 'METADATA_SERVICE_VERSION'
 
+
 async def read_body(request_content):
     byte_array = bytearray()
     while not request_content.at_eof():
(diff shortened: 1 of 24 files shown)


services/utils/__init__.py
```diff
@@ -139,7 +140,7 @@ def __init__(self,
                  password: str = "postgres",
                  database_name: str = "postgres",
                  prefix="MF_METADATA_DB_"):
-        table = str.maketrans({"'": "\'", "`": "\`"})
+        table = str.maketrans({"'": r"\'", "`": r"\`"})
```
saikonen, line 143: has the original translation dictionary ever worked as intended? is there test coverage for this?

msavela (approved):

services/utils/__init__.py
ferras: not sure what you mean. We had a similar piece of code in migration service to deal with passing over the password to goose and it worked since those special characters needed to be escaped

saikonen: From a quick test the first item in the translation dictionary does not seem to escape anything.
translation dict prior to changes:
`{39: "'", 96: '\\`'}`

and after changes:
`{39: "\\'", 96: '\\`'}`

Might the change introduce some unintended behavior or is this the intended escaping behavior?

PR #51, opened 2020-11-16
Failed Run: Duration/finished_at from last known artifact

msavela:
Failed run will now use last known artifact to populate duration/finished_at fields if available.

This PR will increase query cost limits:

```
Before:
> Limit  (cost=2254.34..2254.45 rows=1 width=719)

After:
> Limit  (cost=15378.83..15378.98 rows=1 width=719)
```

This is very significant increase. Suggestions for alternative ways of joining this in more optimized manner is greatly appreciated.


msavela:
Closed due to #127

PR #52, opened 2020-11-16
Revert DBConfguration translate change

msavela:


services/utils/__init__.py
```diff
@@ -140,7 +140,8 @@ def __init__(self,
                  password: str = "postgres",
                  database_name: str = "postgres",
                  prefix="MF_METADATA_DB_"):
-        table = str.maketrans({"'": r"\'", "`": r"\`"})
+        # pylint: disable=anomalous-backslash-in-string
+        table = str.maketrans({"'": "\'", "`": "\`"})
```
saikonen, line 144: you can prepend the second translation with `r` as it used to be, this does not change the translation dictionary in any way and pycodestyle check passes.

saikonen (approved):

PR #54, opened 2020-11-18
Task attempt query changes

saikonen:
* adds backward compatible task attempt query
* changes db table triggers to be configured by a separate list of keys instead of primary keys. Adds 'value' as a trigger field to metadata table trigger (to be able to discern which attempt the metadata belongs to in case of `field_name=attempt`).
* changes notify to broadcast task attempts on metadata triggers instead of task_ok artifact (thus providing realtime tracking instead of after completion/failure task attempts for timeline tracking)


services/data/postgres_async_db.py
```diff
@@ -1027,6 +1031,23 @@ async def update_heartbeat(self, flow_id: str, run_id: str, step_name: str,
...
 
+    async def find_records(self, conditions: List[str] = None, values=[], fetch_single=False,
+                           limit: int = 0, offset: int = 0, order: List[str] = None, groups: List[str] = None,
+                           group_limit: int = 10, expanded=False, enable_joins=False,
+                           postprocess: Callable[[DBResponse], DBResponse] = None,
+                           benchmark: bool = False, overwrite_select_from: str = None
+                           ) -> (DBResponse, DBPagination):
+        if enable_joins:
+            overwrite_select_from = "(SELECT *, UNNEST('{0, 1, 2, 3, 4}'::int[]) as attempt_id FROM tasks_v3) as tasks_v3"
+            conditions.append("NOT (attempt_id > 0 AND started_at IS NULL AND task_ok IS NULL)")
```
saikonen, line 1042: This is where the magic happens. from preliminary testing this kind of query performs immensely better than the previous `UNION` approach for generating existing attempt_ids to join over for the tasks. Slight gotcha is that this works for the current metaflow client implementation, where the `@retry(times=n)` has a hard upper limit of 4 retries, and will perform substantially worse the more 'possible attempts' we need to unnest for and discard in the results.

msavela (approved):

PR #61, opened 2020-11-23
UI service feature flags

msavela:
Feature flags can be controlled with environment variables (will affect UI service only):

- `FEATURE_PREFETCH_DISABLE` (defaults to `0`)
  - Disable any cache prefetching (server start and new data received by metadata service)
- `FEATURE_CACHE_DISABLE` (defaults to `0`)
  - Disable cache completely (**including S3**)
- `FEATURE_S3_DISABLE` (defaults to `0`)
  - Disable all S3 functionality, including cache, prefetch and refine (postprocess)
- `FEATURE_REFINE_DISABLE` (defaults to `0`)
  - Disable all data postprocessing (HTTP GET and Websocket realtime handlers)

Enable flags by setting value to `1` (`FEATURE_PREFETCH_DISABLE=1`)


services/ui_backend_service/api/data_refiner.py
```diff
@@ -75,6 +80,9 @@ def __init__(self):
         super().__init__(field_names=["task_ok", "foreach_stack"])
 
     async def postprocess(self, response: DBResponse):
+        if FEATURE_REFINE_DISABLE:
+            return response
+
```
saikonen, line 85: if we skip postprocessing completely with feature flags, then the api responses will contain internal fields for f.ex. tasks due to them having s3 locations for refinement (foreach_var / _task_ok)

services/ui_backend_service/cache/store.py
```diff
@@ -59,17 +61,21 @@ def __init__(self, event_emitter):
 
         # Bind an event handler for when we want to preload artifacts for
         # newly inserted content.
-        self.event_emitter.on("preload-artifacts", self.preload_event_handler)
-        self.event_emitter.on("run-parameters", self.run_parameters_event_handler)
+        if FEATURE_PREFETCH_ENABLE:
+            self.event_emitter.on("preload-artifacts", self.preload_event_handler)
+            self.event_emitter.on("run-parameters", self.run_parameters_event_handler)
```
saikonen, line 66: run-parameters event is produced by a table trigger. If this is skipped, then the timeline view in the UI never receives run parameters as they become available, instead requiring a page refresh.

services/ui_backend_service/api/data_refiner.py
```diff
@@ -86,7 +91,7 @@ def _process(item):
             item['status'] = 'failed' if item['status'] == 'completed' and item['task_ok'] is False else item['status']
             item.pop('task_ok', None)
 
-            if item['foreach_stack']:
+            if item['foreach_stack'] and len(item['foreach_stack']) > 0 and len(item['foreach_stack'][0]) >= 4:
```
saikonen, line 94: If the `_postprocess` is skipped due to a feature flag, the `foreach_stack` will remain as an S3 location string. Is the logic on the length warranted in this case, even if it works as intended?

services/ui_backend_service/cache/store.py
```diff
@@ -2,16 +2,18 @@
 from services.data.postgres_async_db import AsyncPostgresDB
 from services.data.db_utils import translate_run_key
 from pyee import ExecutorEventEmitter
-from metaflow.client.cache.cache_async_client import CacheAsyncClient
+from metaflow.client.cache.cache_async_client import CacheAsyncClient, CacheClient
```
saikonen: why is CacheClient also imported?

saikonen (approved):

PR #67, opened 2020-11-25
UI: Feature flag FEATURE_MODEL_EXPAND.

msavela:
Introduces feature flag `FEATURE_MODEL_EXPAND` (defaults to `0`) for UI service. This should greatly increase query performance due to `run_number` and `task_id` being part of the index.

From this:
```json
{
  "flow_id": "FlowName",
  "run_number": "run_id",
  "task_id": "task_name",
```
to this:
```json
{
  "flow_id": "FlowName",
  "run_number": 123,
  "run_id": "run_id",
  "task_id": 456,
  "task_name": "task_name",
```

Enable expanded models using env variable `FEATURE_MODEL_EXPAND=1`

UI frontend should be compatible with this for the most part.


services/ui_backend_service/cache/store.py
```diff
@@ -101,7 +101,7 @@ async def get_recent_run_numbers(self):
             values=[int(round(time.time() * 1000)) - (int(METAFLOW_ARTIFACT_PREFETCH_RUNS_SINCE) * 1000)],
             order=['ts_epoch DESC'],
             limit=METAFLOW_ARTIFACT_PREFETCH_RUNS_LIMIT,
-            expanded=True
+            expanded=FEATURE_MODEL_EXPAND
```
saikonen, line 104: wasn't this used only internally for preloading data for runs? in this case it could be left to always expanded, so it uses run numbers internally.

msavela: True, will change it back

saikonen (approved):

PR #76, opened 2020-12-14
UI feature flags finalization

msavela:
Removed unnecessary UI feature flags:

- `FEATURE_MODEL_EXPAND` - UI is adapted to this mode already and is no longer necessary
- `FEATURE_JOIN_DISABLE` - Most functionality will break if joins are disabled
- `FEATURE_DB_TRIGGER_DISABLE` - Handled by `create_triggers` in `db._init()`

Added new environment variable:

- `DB_TRIGGER_CREATE=1` - Will force creation of table triggers, works with any service using `services/data/postgres_async_db.py` and is enabled by default only on UI service (disable with `DB_TRIGGER_CREATE=0`)


services/data/postgres_async_db.py
```diff
@@ -53,7 +52,7 @@ def __init__(self, name='global'):
         tables.append(self.metadata_table_postgres)
         self.tables = tables
 
-    async def _init(self, db_conf: DBConfiguration):
+    async def _init(self, db_conf: DBConfiguration, create_triggers=False):
```
saikonen, line 55: will having this default to false cause problems with some deployments? I remember talks about running the ui_backend on a read replica, which would mean that the metadata service would be responsible for creating the triggers on the write master?

msavela: Excellent point, might be best to put this behind feature flag after all.

services/data/postgres_async_db.py
```diff
@@ -16,12 +16,14 @@
...
 AIOPG_ECHO = os.environ.get("AIOPG_ECHO", 0) == "1"
-FEATURE_DB_TRIGGER_DISABLE = os.environ.get("FEATURE_DB_TRIGGER_DISABLE", 0) == "1"
-FEATURE_JOIN_DISABLE = os.environ.get("FEATURE_JOIN_DISABLE", 0) == "1"
 
 WAIT_TIME = 10
 OLD_RUN_FAILURE_CUTOFF_TIME = 60 * 60 * 24 * 1000 * 14  # 2 weeks (in milliseconds)
 
+# Create database triggers automatically, disabled by default
+# Enable with env variable `DB_TRIGGER_CREATE=1`
+DB_TRIGGER_CREATE = os.environ.get("DB_TRIGGER_CREATE", 0) == "1"
```
saikonen, line 25: default still remains 'false' so metadata service needs to specifically have the env var set so it creates triggers when running on a write master? Not an issue, just a configuration hurdle that should probably be noted.

saikonen (approved):

PR #84, opened 2021-01-12
Separate data refinement from basic task routes

saikonen:
* separate slow postprocessing from basic task routes. Introduces `postprocess=true` query parameter for task routes to control postprocessing
* update related integration tests
* introduces an 'unknown' status for tasks for cases where status can not be inferred from database content. This is used as a placeholder until S3 content finishes loading (with the separate refined request). Only relevant for old runs that did not produce enough status-related metadata


services/ui_backend_service/api/ws.py
```diff
@@ -108,6 +108,8 @@ async def subscribe_to(self, ws, uuid: str, resource: str, since: int):
         await self.unsubscribe_from(ws, uuid)
 
         # Create new subscription
+        # TODO: this is only for testing purposes, a quick and dirty cleanup of invalid query params!
+        resource = resource.replace("&postprocess=false", "")
```
msavela, line 112: Do we keep this? I assume `?postprocess=` will not work

saikonen: removed for now, UI takes care of passing proper params. the subscription filter params handling needs a more thorough rework in another PR

msavela (approved):

PR #89, opened 2021-01-14
Increase heartbeat_ts threshold limit

saikonen:
- increases heartbeat threshold to three times the heartbeat interval.


savingoyal:
Can we increase it to 6x to begin with?

msavela (approved):

PR #93, opened 2021-01-22
Fix db error reporting

peterwilton-ttam:
This is an apparent fix of the bug mentioned [here](https://gitter.im/metaflow_org/community?at=5fddaea4c746c6431cc361a4). This bug obfuscates any error response stored by `self._async_table.get_artifact_in_task(...)` in `DBResponse.body`.

Basically, when there is a generic DB error, `self._async_table.get_artifact_in_task(...)` returns a `DBResponse` with `.body` attribute of type string (containing the error message) and a `.response_code` of `500`, indicating an error. The string-type body is set in `db_utils.aiopg_exception_handling`.

`ArtifactsApi._filter_artifacts_by_attempt_id(...)` expects `artifacts.body` to be an iterable, which is what `get_artifact_in_task` returns when no error has occurred (and `fetch_single==False`, which is the case in `get_artifact_in_task`). This fix avoids the call to `ArtifactsApi._filter_artifacts_by_attempt_id(...)` when `DBResponse.response_code` is not `200`.

### DIFF
--- a/services/metadata_service/api/artifact.py
+++ b/services/metadata_service/api/artifact.py
@@ -137,7 +137,7 @@ class ArtificatsApi(object):
         )
 
         filtered_body = ArtificatsApi._filter_artifacts_by_attempt_id(
-            artifacts.body)
+            artifacts)
         return web.Response(
             status=artifacts.response_code, body=json.dumps(filtered_body)
         )
@@ -181,7 +181,7 @@ class ArtificatsApi(object):
         )
 
         filtered_body = ArtificatsApi._filter_artifacts_by_attempt_id(
-            artifacts.body)
+            artifacts)
         return web.Response(
             status=artifacts.response_code, body=json.dumps(filtered_body)
         )
@@ -216,7 +216,7 @@ class ArtificatsApi(object):
 
         artifacts = await self._async_table.get_artifacts_in_runs(flow_name, run_number)
         filtered_body = ArtificatsApi._filter_artifacts_by_attempt_id(
-            artifacts.body)
+            artifacts)
         return web.Response(
             status=artifacts.response_code, body=json.dumps(filtered_body)
         )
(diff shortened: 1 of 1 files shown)


peterwilton-ttam:
Here is the full comment on gitter for thoroughness:

------

I'm trying to take a look at the artifacts associated with one of our SFN executions but when trying to call:

```
Step(...).task.data
```
We run into the following error:

```
ServiceException: Metadata request (/flows/{Flow}/runs/{Run}/steps/start/tasks/{Task}/artifacts) failed (code 500): 500 Internal Server Error
```
Looking at the logs from our metadata service, I see the following error:
```
Traceback (most recent call last):
    File "/opt/latest/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 418, in start
        resp = await task
    File "/opt/latest/lib/python3.7/site-packages/aiohttp/web_app.py", line 458, in _handle
        resp = await handler(request)
    File "/opt/latest/lib/python3.7/site-packages/services/metadata_service/api/artifact.py", line 140, in get_artifacts_by_task
        artifacts.body)
    File "/opt/latest/lib/python3.7/site-packages/services/metadata_service/api/artifact.py", line 355, in _filter_artifacts_by_attempt_id
        attempt_id = ArtificatsApi._get_latest_attempt_id(artifacts)
    File "/opt/latest/lib/python3.7/site-packages/services/metadata_service/api/artifact.py", line 349, in _get_latest_attempt_id
        if artifact['attempt_id'] > attempt_id:
TypeError: string indices must be integers
```
From what I can tell, this can only occur when the result of [this call](https://github.com/Netflix/metaflow-service/blob/b656d42cfb946b5ec64ef15b4f5e86a1505a4e90/services/data/postgres_async_db.py#L112-L156) returns a `DBResponse` object with a body of type string. And it looks like this would only happen in the case of error handling. Unfortunately it looks like the underlying error message is obfuscated by this error.

PR #94, opened 2021-01-25
UI backend service structure cleanup

saikonen:
- clean up project structure to start utilizing modules more.
- refactors problematic singleton pattern from CacheStore that does not play well with modules.


services/ui_backend_service/api/artifactsearch.py
```diff
@@ -2,22 +2,21 @@
...
         app.router.add_route(
             "GET", "/flows/{flow_id}/runs/{run_number}/search", self.get_run_tasks
         )
         self._artifact_table = self.db.artifact_table_postgres
         self._run_table = self.db.run_table_postgres
-        self._artifact_store = CacheStore().artifact_cache
+        self._artifact_store = getattr(cache, "artifact_cache", None)
 
-    @handle_exceptions
+    @ handle_exceptions
```
msavela, line 19: Extra whitespace

saikonen: fixed, good catch!

msavela (approved):

PR #97, opened 2021-01-27
Remove pagination last page due to performance impact

msavela:
Remove `COUNT(*) OVER() AS count_total` from API queries due to performance impact.

### DIFF
--- a/services/ui_backend_service/api/utils.py
+++ b/services/ui_backend_service/api/utils.py
@@ -25,12 +25,12 @@ def format_response(request: web.BaseRequest, db_response: DBResponse) -> (int, 
     return db_response.response_code, response_object
 
 
-def format_response_list(request: web.BaseRequest, db_response: DBResponse, page: int, lastPage: int) -> (int, Dict):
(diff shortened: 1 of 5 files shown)


services/ui_backend_service/api/utils.py
```diff
@@ -25,12 +25,12 @@ def format_response(request: web.BaseRequest, db_response: DBResponse) -> (int,
...
 
 
-def format_response_list(request: web.BaseRequest, db_response: DBResponse, page: int, lastPage: int) -> (int, Dict):
+def format_response_list(request: web.BaseRequest, db_response: DBResponse, page: int) -> (int, Dict):
     query = {}
     for key in request.query:
         query[key] = request.query.get(key)
 
-    nextPage = min(page + 1, lastPage)
+    nextPage = page + 1
```
saikonen, line 33: could this have a check if the DBResponse body is empty, to instead not provide nextpage at all, or do we always want to provide a link?

saikonen (approved):

services/ui_backend_service/api/utils.py
msavela: Excellent point, will update this.

PR #106, opened 2021-02-02
Various order query indexes

msavela:
Add following indexes to table `runs_v3`:

- `runs_v3_idx_ts_epoch`
- `runs_v3_idx_gin_tags_combined`
- `runs_v3_idx_flow_id_asc_ts_epoch_desc`
- `runs_v3_idx_user_asc_ts_epoch_desc`

This only affects UI service performance while sorting and filtering list of runs.


services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+CREATE INDEX runs_v3_idx_flow_id ON runs_v3 (flow_id);
```
msavela, line 5: NOTE: This might not be necessary

services/migration_service/api/__init__.py
```diff
@@ -4,7 +4,8 @@
     '0': 'v_1_0_1',
     '1': 'v_1_0_1',
     '20200603104139': 'latest',
-    '20201002000616': 'latest'
+    '20201002000616': 'latest',
+    '20210202145952': 'latest'
```
ferras, line 8: you can't have multiple versions pointing to latest, this may cause undefined behavior

msavela: Good point, https://github.com/Netflix/metaflow-service/pull/30 this PR might have the same conflict as well.

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,62 @@
...
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+-- Others
+
+CREATE INDEX runs_v3_idx_ts_epoch ON runs_v3 (ts_epoch);
+
+CREATE INDEX runs_v3_idx_gin_tags ON runs_v3 USING gin (tags, system_tags);
+CREATE INDEX runs_v3_idx_gin_tags_combined ON runs_v3 USING gin ((tags || system_tags));
+CREATE INDEX runs_v3_idx_gin_tags_combined_tsvector ON runs_v3 USING gin (to_tsvector('simple', (tags||system_tags)));
```
msavela, line 11: This needs to be removed due to PostgreSQL 9 requirement

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,60 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+-- Others
+
+CREATE INDEX runs_v3_idx_ts_epoch ON runs_v3 (ts_epoch);
```
ferras, line 7: note that creating an index using the goose script will lock the table and cause some amount of downtime. The runtable is relatively small so not really a major issue however we are working on an alternative way to add indices concurrently

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,60 @@
...
+
+CREATE INDEX runs_v3_idx_gin_tags ON runs_v3 USING gin (tags, system_tags);
```
ferras, line 9: when creating an index add `IF NOT EXISTS` to prevent the migration scripts from getting stuck if a user already manually added the indices

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,60 @@
...
+    (CASE
+        WHEN system_tags ? ('user:' || user_name)
+        THEN user_name
+        ELSE NULL
+    END) DESC, ts_epoch DESC
+);
+
+-- run
+
+CREATE INDEX runs_v3_idx_run_asc ON runs_v3 (COALESCE(run_id, run_number::text) ASC);
```
ferras, line 36: these are redundant

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,60 @@
...
+
+CREATE INDEX runs_v3_idx_ts_epoch ON runs_v3 (ts_epoch);
+
+CREATE INDEX runs_v3_idx_gin_tags ON runs_v3 USING gin (tags, system_tags);
+CREATE INDEX runs_v3_idx_gin_tags_combined ON runs_v3 USING gin ((tags || system_tags));
+
+-- flow_id + ts_epoch
+
+CREATE INDEX runs_v3_idx_flow_id_asc_ts_epoch_desc ON runs_v3 (flow_id ASC, ts_epoch DESC);
+CREATE INDEX runs_v3_idx_flow_id_desc_ts_epoch_desc ON runs_v3 (flow_id DESC, ts_epoch DESC);
```
ferras, line 15: depending on what you intend on using this index for you can possibly combine these into a single index by reordering to
CREATE INDEX runs_v3_idx_ts_epoch_desc_flow_id ON runs_v3 (ts_epoch DESC, flow_id);


Note an index already exists on flow_id

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
saikonen: good point, forgot about this :ok_hand:

Situation can also happen if metadata service ran the create tables command and created indices along with it, before migration service was used (can and will happen a lot during local development).

services/migration_service/migration_files/20210202145952_add_runs_idx_ts_epoch_flow_id.sql
```diff
@@ -0,0 +1,60 @@
...
+
+-- user && ts_epoch
+
+CREATE INDEX runs_v3_idx_user_asc_ts_epoch_desc ON runs_v3 (
```
ferras, line 19: can you remind me again why it's not possible to just create an index on the user_name column without parsing the tags?
If the user_name column isn't being properly populated then that should be fixed

saikonen: from what I recall this was primarily implemented for the query for grouping by user. When a flow is run on SFN, the user_name is set to 'SFN', which slams that result set in between actual users, instead of what is preferred (NULLS LAST). @msavela have any addional info?

saikonen (approved):

PR #110, opened 2021-02-04
Fix integration tests getting stuck.

msavela:
PostgreSQL connection kept timing out (after 60s x3) and the integration job appeared "stuck" due to timeout+retry loop.

For some reason `localhost` stopped working and connecting to `127.0.0.1` seems to fix the problem.

### DIFF
--- a/services/utils/__init__.py
+++ b/services/utils/__init__.py
@@ -146,8 +146,10 @@ class DBConfiguration(object):
 
     # aiopg default pool sizes
     # https://aiopg.readthedocs.io/en/stable/_modules/aiopg/pool.html#create_pool
-    pool_min: int = 1
-    pool_max: int = 10
+    pool_min: int = None  # aiopg default: 1
+    pool_max: int = None  # aiopg default: 10
+
+    timeout: int = None  # aiopg default: 60 (seconds)
 
     _dsn: str = None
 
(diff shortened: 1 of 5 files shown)


services/utils/__init__.py
```diff
@@ -174,6 +177,8 @@ def __init__(self,
         self.pool_min = int(os.environ.get(prefix + "POOL_MIN", self.pool_min))
         self.pool_max = int(os.environ.get(prefix + "POOL_MAX", self.pool_max))
 
+        self.timeout = int(os.environ.get(prefix + "TIMEOUT", self.timeout))
+
```
saikonen, line 181: should this default to `timeout` instead of `self.timeout` or does it work as intended?

saikonen (approved):

services/utils/__init__.py
```diff
@@ -171,8 +174,10 @@ def __init__(self,
         self.database_name = os.environ.get(
             prefix + "NAME", database_name).translate(table)
 
-        self.pool_min = int(os.environ.get(prefix + "POOL_MIN", self.pool_min))
-        self.pool_max = int(os.environ.get(prefix + "POOL_MAX", self.pool_max))
+        self.pool_min = int(os.environ.get(prefix + "POOL_MIN", pool_min))
+        self.pool_max = int(os.environ.get(prefix + "POOL_MAX", pool_max))
```
saikonen, line 178: init doesn't seem to take pool_min/pool_max as inputs currently so these are unassigned

PR #118, opened 2021-02-09
Performance improvement for specific tag filtering scenario.

msavela:


services/ui_backend_service/api/run.py
```diff
@@ -93,13 +93,41 @@ async def get_all_runs(self, request):
...
+            if len(order) > 0:
+                overwrite_table_name = "(SELECT * FROM runs_v3 {order_by}) AS runs_v3".format(
+                    order_by="ORDER BY {}".format(", ".join(order)) if order else ""
+                )
+
+                return await find_records(request,
+                                          self._async_table,
+                                          initial_conditions=[],
+                                          initial_values=[],
+                                          allowed_order=[],  # This is already handled by above `pagination_query`
```
saikonen, line 117: tiny nit. are the empty lists necessary, doesn't the find_records default to these?

services/ui_backend_service/api/run.py
```diff
@@ -93,13 +93,41 @@ async def get_all_runs(self, request):
...
+        # JSONB tag filters combined with `ORDER BY` causes performance impact
+        # due to lack of pg statistics on JSONB fields. To battle this, first execute
+        # subquery of ordered list from runs_v3 table in and filter by tags on outer query.
+        # This needs more research in the future to further improve performance.
+        builtin_conditions, _ = builtin_conditions_query(request)
+        has_tag_filter = len([s for s in builtin_conditions if 'tags||system_tags' in s]) > 0
+        if has_tag_filter:
+            _, _, _, order, _, _ = pagination_query(request, allowed_order=allowed_order, allowed_group=[])
+            if len(order) > 0:
+                overwrite_table_name = "(SELECT * FROM runs_v3 {order_by}) AS runs_v3".format(
```
saikonen, line 109: might be just me, but `overwrite_table_name` doesn't clearly convey that this overrides the `FROM` in the produced query, maybe something like `overwrite_select_from` ? Not a blocker for merging.

saikonen (approved):

PR #122, opened 2021-02-12
Optimized tag filters for flow_id, ts_epoch order

msavela:
Use optimized tag filter query when order key is `flow_id` and/or `ts_epoch`. Otherwise use the non-performant query that supports computed columns from joins.

### DIFF
--- a/services/ui_backend_service/api/run.py
+++ b/services/ui_backend_service/api/run.py
@@ -104,10 +104,16 @@ class RunApi(object):
         builtin_conditions, _ = builtin_conditions_query(request)
         has_tag_filter = len([s for s in builtin_conditions if 'tags||system_tags' in s]) > 0
         if has_tag_filter:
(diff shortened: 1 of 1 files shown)


saikonen (approved):

services/ui_backend_service/api/run.py
```diff
@@ -104,10 +104,16 @@ async def get_all_runs(self, request):
         builtin_conditions, _ = builtin_conditions_query(request)
         has_tag_filter = len([s for s in builtin_conditions if 'tags||system_tags' in s]) > 0
         if has_tag_filter:
-            _, _, _, order, _, _ = pagination_query(request, allowed_order=allowed_order, allowed_group=[])
-            if len(order) > 0:
+            allowed_optimized_order = ['flow_id', 'ts_epoch']
+            allowed_unoptimized_order = [o for o in allowed_group if o not in ['flow_id', 'ts_epoch']]
```
saikonen, line 108: could instead be `if o not in allowed_optimized_order` in case the above gets changed?

PR #124, opened 2021-02-18
Custom UI backend data adapter

saikonen:
Separates the database adapter as much as possible from the common `data` module, due to having very different access requirements between metadata and ui services. Uses the shared module as a single source of truth for schema specific things (creating tables and triggers, tracking schema version)

- [x] separate data adapter to ui_service specific module
- [x] get rid of singleton pattern for data adapter
- [x] explicitly require initialized db for all components depending on one.
- [x] clean up `data.postgres_async_db`, removing ui specific changes as unnecessary for a shared component.


services/data/postgres_async_db.py
```diff
@@ -205,165 +164,55 @@ async def get_records(self, filter_dict={}, fetch_single=False,
...
 
-                records = await cur.fetchall()
-                rows = []
-                for record in records:
-                    rows.append(record[0])
-                return "\n".join(rows)
-        except (Exception, psycopg2.DatabaseError):
-            self.db.logger.exception("Query Benchmarking failed")
-            return None
+        select_sql = sql_template.format(
```
msavela, line 189: Duplicate `select_sql`

services/ui_backend_service/data/db/tables/artifact.py
```diff
@@ -0,0 +1,67 @@
...
+    table_name = MetadataArtifactTable.table_name
+    task_table_name = AsyncTaskTablePostgres.table_name
+    ordering = ["attempt_id DESC"]
+    keys = MetadataArtifactTable.keys
+    primary_keys = MetadataArtifactTable.primary_keys
+    trigger_keys = MetadataArtifactTable.trigger_keys
+    select_columns = keys
+    _command = MetadataArtifactTable._command
+
+    async def get_artifacts_in_runs(self, flow_id: str, run_id: int):
```
msavela, line 21: Should we get rid of these, since most of the UI service relies on `find_records`

services/data/postgres_async_db.py
saikonen: good catch.

services/ui_backend_service/data/db/tables/artifact.py
msavela: This comment applies to most of the `tables/*.py` `get_*` functions

saikonen: very true, will clean these up.

services/ui_backend_service/data/db/tables/run.py
```diff
@@ -0,0 +1,113 @@
+from .base import AsyncPostgresTable, HEARTBEAT_THRESHOLD, OLD_RUN_FAILURE_CUTOFF_TIME, WAIT_TIME
+from .flow import AsyncFlowTablePostgres
+from ..models import RunRow
+from services.data.db_utils import DBResponse, translate_run_key
+# use schema constants from the .data module to keep things consistent
+from services.data.postgres_async_db import AsyncRunTablePostgres as MetadataRunTable
+import json
```
msavela, line 7: Some unused imports here

msavela (approved):

PR #127, opened 2021-02-23
Feature/correct run failure detection

saikonen:
an assortment of fixes related to runs.
- fix finished_at and duration calculations for runs with no heartbeat to happen based on the latest artifact timestamp. Applies for failed runs.
- correctly mark runs without heartbeat as 'failed' when a task attempt fails finally (a new attempt has not started within the threshold period and current attempt is failed). Fixes runs stuck in 'running' state when no end step has been produced and worker has stopped.

some benchmark highlights from an environment, with the following request:
`/api/runs?_order=-ts_epoch&_limit=15&_group_limit=16&_page=1&benchmark=true`

UI branch 
`
Limit  (cost=1.38..221.05 rows=15 width=372) (actual time=0.058..0.346 rows=15 loops=1)
Planning Time: 1.073 ms
Execution Time: 0.424 ms
`

current branch:
`
Limit  (cost=21.51..347.12 rows=15 width=372) (actual time=0.137..0.988 rows=15 loops=1)
Planning Time: 0.513 ms
Execution Time: 1.080 ms
`


msavela (approved):

msavela (approved):

msavela (commented):
Update PR description to reflect recent changes (EXPLAIN ANALYZE etc.)

services/ui_backend_service/data/db/tables/base.py
```diff
@@ -14,6 +14,9 @@
 HEARTBEAT_THRESHOLD = int(os.environ.get("HEARTBEAT_THRESHOLD", WAIT_TIME * 6))
 OLD_RUN_FAILURE_CUTOFF_TIME = int(os.environ.get("OLD_RUN_FAILURE_CUTOFF_TIME", 60 * 60 * 24 * 1000 * 14))  # default 2 weeks (in milliseconds)
 
+# Expected Step-functions Scheduler delay between task starts in milliseconds. Default two minutes.
+SCHEDULER_DELAY = int(os.environ.get("SCHEDULER_DELAY", 60 * 2 * 1000))
+
```
saikonen, line 19: scheduler delay can be in days, so this is not really a workable solution for even most of the cases. Scrap the scheduler delay for now but push the other changes forward.

PR #132, opened 2021-03-12
fix: Enable triggers always and use periodic ping for NOTIFY.

msavela:
This PR is a workaround for PostgreSQL bug in logical replication mode where notifications via triggers were not delivered properly.

Solution here is to manually `NOTIFY ping` channel in intervals so that the main `NOTIFY notify` channel "buffer" is delivered along with the ping channel. It doesn't matter which channel is notified, important thing here is that the notification originates from non-replication source. This bug is present in at least PostgreSQL versions 10 - 13.

`NOTIFY ping` is executed every 1 second, so all replication notifications will be delivered in 1 second intervals in this special use-case.

Read more about the bug here: https://www.postgresql.org/message-id/153243441449.1404.2274116228506175596%40wrigleys.postgresql.org

This PR also enables WAL level `logical` on development mode for any future replication tests.

### DIFF
--- a/docker-compose.development.yml
+++ b/docker-compose.development.yml
@@ -43,6 +43,7 @@ services:
       - FEATURE_HEARTBEAT_DISABLE=0
       - FEATURE_DB_LISTEN_DISABLE=0
       - CUSTOM_QUICKLINKS=$CUSTOM_QUICKLINKS
+      - GA_TRACKING_ID=none
     depends_on:
       - db
   metadata:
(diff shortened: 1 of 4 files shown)


docker-compose.development.yml
```diff
@@ -43,6 +43,7 @@ services:
       - FEATURE_HEARTBEAT_DISABLE=0
       - FEATURE_DB_LISTEN_DISABLE=0
       - CUSTOM_QUICKLINKS=$CUSTOM_QUICKLINKS
+      - GA_TRACKING_ID=none
```
saikonen, line 46: what is this used for?

msavela: This is a Google Analytics tracking id that is forwarded to the frontend, it's not necessary here but it might be good to have all available envs listed here, what do you think?

saikonen: Ah, sounds like a good addition yes. Was just a bit baffled as it is not directly related to the PR in question.

saikonen (approved):

PR #136, opened 2021-03-17
feat: Autocomplete endpoints for UI

rsanteri:
Draft for autocomplete

Autocomplete endpoints:
/autocomplete/tags
/autocomplete/flows
/autocomplete/flows/:flowname/runs
/autocomplete/flows/:flowname/runs/:runid/step

Result for tags endpoint is cached within class and refreshed every 5min since this request might be rather slow to database.

### DIFF
--- a/services/ui_backend_service/api/autocomplete.py
+++ b/services/ui_backend_service/api/autocomplete.py
@@ -2,6 +2,9 @@ from services.utils import handle_exceptions, web_response
 import asyncio
 import threading
 
+# Interval for refetching tags from database. Tags data is cached since requesting all of them might take a while.
+TAGS_FILL_INTERVAL = 300
+
 
 class AutoCompleteApi(object):
     # Cache tags so we don't have to request DB everytime
@@ -17,16 +20,21 @@ class AutoCompleteApi(object):
         app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs/{run_id}/steps", self.get_steps_for_run)
         self._async_table = self.db.run_table_postgres
         self.loop = asyncio.get_event_loop()
-        self.loop.create_task(self.setup_tags())
+        self.loop.create_task(self.fill_tags_cache())
 
     @handle_exceptions
-    async def setup_tags(self):
-        db_response = await self._async_table.get_tags()
-
-        if db_response.response_code == 200:
-            self.tags = db_response.body
-        # Check tags again in 5 minutes
-        self.loop.call_later(300, lambda x: self.loop.create_task(self.setup_tags()), self)
+    async def fill_tags_cache(self):
+        '''
+        Async task that fill tags cache every 5minutes. Database query might take a while
+        so its better to cache the result.
+        '''
+        while True:
+            # Get all tags taht are mentioned in runs table
+            db_response = await self._async_table.get_tags()
+            if db_response.response_code == 200:
+                self.tags = db_response.body
+            # Check tags again after some sleep
+            await asyncio.sleep(TAGS_FILL_INTERVAL)
 
     @handle_exceptions
     async def get_all_tags(self, request):
--- a/services/ui_backend_service/api/autocomplete.py
+++ b/services/ui_backend_service/api/autocomplete.py
@@ -1,0 +1,53 @@
+from services.utils import handle_exceptions, web_response
+import asyncio
+
+class AutoCompleteApi(object):
+
+    def __init__(self, app, db):
+        self.db = db
+        # Cached resources
+        app.router.add_route("GET", "/autocomplete/tags", self.get_all_tags)
+        app.router.add_route("GET", "/autocomplete/flows", self.get_all_flows)
+        # Non-cached resources
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs", self.get_runs_for_flow)
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs/{run_id}/steps", self.get_steps_for_run)
+        self._async_table = self.db.run_table_postgres
+        self.loop = asyncio.get_event_loop()
+        self.loop.create_task(self.setup_tags())
+        self.loop.create_task(self.setup_flows())
+
+    @handle_exceptions
+    async def setup_tags(self):
+        db_response = await self._async_table.get_tags()
+        self.tags = db_response.body
+
+    @handle_exceptions
+    async def setup_flows(self):
+        db_response = await self._async_table.get_field_from(field="flow_id", table="flows_v3")
+        self.flows = db_response.body
+
+    @handle_exceptions
+    async def get_all_tags(self, request):
+        return web_response(200, self.tags)
+
+    @handle_exceptions
+    async def get_all_flows(self, request):
+        return web_response(200, self.flows)
+
+    @handle_exceptions
+    async def get_runs_for_flow(self, request):
+        flowid = request.match_info.get("flow_id")
+        sql_conditions = ["flow_id=%s"]
+        db_response = await self._async_table.get_field_from(field="COALESCE(run_id, run_number::text)", table="runs_v3", conditions=sql_conditions, values=[flowid])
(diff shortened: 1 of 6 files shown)


services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+        self.db = db
+        # Cached resources
+        app.router.add_route("GET", "/autocomplete/tags", self.get_all_tags)
+        # Non-cached resources
+        app.router.add_route("GET", "/autocomplete/flows", self.get_all_flows)
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs", self.get_runs_for_flow)
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs/{run_id}/steps", self.get_steps_for_run)
+        self._async_table = self.db.run_table_postgres
+        self.loop = asyncio.get_event_loop()
+        self.loop.create_task(self.setup_tags())
```
saikonen, line 20: recursively continuing tasks works as well, but a simpler solution might be to have a poller task that uses asyncio.sleep at the end? see `heartbeat_monitor.py#check_heartbeats` for an example.

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @handle_exceptions
+    async def setup_tags(self):
```
saikonen, line 23: maybe slightly confusing naming, is not immediately apparent that the function does fetch&store&sleep looping.

services/ui_backend_service/data/db/tables/base.py
```diff
@@ -256,3 +256,29 @@ async def get_tags(self):
...
+
+                items = []
+                records = await cur.fetchall()
+                for record in records:
+                    items += record
+                cur.close()
+                return DBResponse(response_code=200, body=items)
+        except (Exception, psycopg2.DatabaseError) as error:
+            self.db.logger.exception("Exception occured")
+            return aiopg_exception_handling(error)
```
saikonen, line 284: instead of re-implementing the execution, could you use the `execute_sql` function that should be present for this scope?

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+        produces:
+        - application/json
+        responses:
+            "200":
+                description: Returns string list of all flow id's
+                type: array
+                items:
+                    type: string
+        """
+        db_response = await self._async_table.get_field_from(field="flow_id", table="flows_v3")
```
saikonen, line 65: table names should be pulled directly from the db class instead of hardcoding, as these are subject to change whenever. So for this case rather use `table=self._async_table.table_name`

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+        responses:
+            "200":
+                description: Returns string list of run ids
+                type: array
+                items:
+                    type: string
+        """
+        flowid = request.match_info.get("flow_id")
+        sql_conditions = ["flow_id=%s"]
+        db_response = await self._async_table.get_field_from(field="COALESCE(run_id, run_number::text)", table="runs_v3", conditions=sql_conditions, values=[flowid])
```
saikonen, line 86: reiterating the other comment about static table names :)

saikonen (commented):
some integration tests for the new api would also be appreciated to cover basic (expected) functionality.

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class AutoCompleteApi(object):
+    # Cache tags so we don't have to request DB everytime
+    tags = []
+
+    def __init__(self, app, db):
+        self.db = db
+        # Cached resources
+        app.router.add_route("GET", "/autocomplete/tags", self.get_all_tags)
```
msavela, line 13: Should we also include users here (cached) so that we have all the Home requirements covered?

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,110 @@
...
+        # Non-cached resources
+        app.router.add_route("GET", "/autocomplete/flows", self.get_all_flows)
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs", self.get_runs_for_flow)
+        app.router.add_route("GET", "/autocomplete/flows/{flow_id}/runs/{run_id}/steps", self.get_steps_for_run)
```
saikonen, line 17: opinions on routing, feel free to disagree: should we stick to the original restful paths and instead have the `autocomplete` as a trailing resource? i.e. `/flows/{flow_id}/runs/autocomplete` as it is a query that is targeted at a resourcepath, instead of "autocomplete" being a distinct resource?

services/ui_backend_service/data/db/tables/base.py
```diff
@@ -256,3 +256,29 @@ async def get_tags(self):
         except (Exception, psycopg2.DatabaseError) as error:
             self.db.logger.exception("Exception occured")
             return aiopg_exception_handling(error)
+
+    async def get_field_from(self, field: str, table: str, conditions: List[str] = [], values: List[str] = []):
+        sql_template = "SELECT {field_name} FROM {table_name} {conditions}"
```
msavela, line 261: `DISTINCT` might be a good idea here since

services/ui_backend_service/data/db/tables/base.py
rsanteri: Hmm, it seems that execute_sql tries to serialize stuff before returning. Here i just want list of strings.

rsanteri: I'm not currently utilizing those table objects here. I guess i should

services/ui_backend_service/api/autocomplete.py
rsanteri: Currently i'm using tags for users field as well. Do you think we should get separate these endpoints? Good thing abotu using tags is that we send only 2 initial requests on frontpage, flows and tags where tags includes project, user and tags field

services/ui_backend_service/api/autocomplete.py
rsanteri: /flows/autocomplete does clash with `"/flows/{flow_id}"` etc but maybe i can prioritize autocomplete endpoints first

saikonen: mm that's true.. and can't have a flow with the name autocomplete then. good point

services/ui_backend_service/data/db/tables/base.py
saikonen: ah, true you want a simple list as a return. Disregard this then, should be fine as is.

services/ui_backend_service/data/db/tables/base.py
```diff
@@ -256,3 +256,29 @@ async def get_tags(self):
         except (Exception, psycopg2.DatabaseError) as error:
             self.db.logger.exception("Exception occured")
             return aiopg_exception_handling(error)
+
+    async def get_field_from(self, field: str, table: str, conditions: List[str] = [], values: List[str] = []):
```
saikonen: not sure the table is necessary as a parameter here? the function gets inherited by all table classes, so `self.table_name` should be available. Only change would be that you specifically call `.get_fields` on the table instance you want to perform it on

rsanteri: Yep, Removed

services/ui_backend_service/api/autocomplete.py
```diff
@@ -0,0 +1,129 @@
+from services.utils import handle_exceptions, web_response
+from services.data.db_utils import translate_run_key
+import asyncio
+import threading
```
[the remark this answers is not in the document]
rsanteri: Yep

saikonen (approved):

PR #138, opened 2021-03-18
refactor: Introduce ParameterRefiner to refine Run parameters.

msavela:
- Introduces new `ParameterRefiner`
- `ParameterRefiner` is now used in `ArtifactCacheStore`
   - `GET /flows/{flow_id}/runs/{run_number}/parameters` uses `ArtifactCacheStore` internally
- `CacheStore` table -specific logic moved under `ui_service_backend/data/db/tables/`.

### DIFF
--- a/services/ui_backend_service/api/task.py
+++ b/services/ui_backend_service/api/task.py
@@ -28,7 +28,7 @@ class TaskApi(object):
             self.get_task_attempts,
         )
         self._async_table = self.db.task_table_postgres
-        self.refiner = TaskRefiner(cache=cache.artifact_cache if cache else None)
+        self.refiner = TaskRefiner(cache=cache.artifact_cache) if cache else None
 
     @handle_exceptions
     async def get_run_tasks(self, request):
--- a/services/ui_backend_service/api/task.py
+++ b/services/ui_backend_service/api/task.py
@@ -28,7 +28,7 @@ class TaskApi(object):
             self.get_task_attempts,
         )
         self._async_table = self.db.task_table_postgres
-        self.refiner = TaskRefiner(cache=cache)
+        self.refiner = TaskRefiner(cache=cache.artifact_cache if cache else None)
 
     @handle_exceptions
     async def get_run_tasks(self, request):
--- a/services/ui_backend_service/data/cache/store.py
+++ b/services/ui_backend_service/data/cache/store.py
@@ -3,7 +3,7 @@ from pyee import AsyncIOEventEmitter
 from metaflow.client.cache.cache_async_client import CacheAsyncClient
 from .search_artifacts_action import SearchArtifacts
 from .get_artifacts_action import GetArtifacts
-from ..refiner.parameter_refiner import ParameterRefiner, GetParametersFailed
+from ..refiner import ParameterRefiner
 import asyncio
 import time
 import os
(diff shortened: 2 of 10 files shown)


services/ui_backend_service/api/task.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, app, db, cache=None):
             self.get_task_attempts,
         )
         self._async_table = self.db.task_table_postgres
-        self.refiner = TaskRefiner(cache=cache)
+        self.refiner = TaskRefiner(cache=cache.artifact_cache if cache else None)
```
saikonen, line 31: should the else None be outside the init perhaps? as is the case in `ws.py`

services/ui_backend_service/data/cache/store.py
```diff
@@ -1,9 +1,9 @@
 from .generate_dag_action import GenerateDag
-from services.data.db_utils import translate_run_key
 from pyee import AsyncIOEventEmitter
 from metaflow.client.cache.cache_async_client import CacheAsyncClient
 from .search_artifacts_action import SearchArtifacts
 from .get_artifacts_action import GetArtifacts
+from ..refiner.parameter_refiner import ParameterRefiner, GetParametersFailed
```
saikonen, line 6: exception is now unused in this file, is it not?

saikonen: also paramrefiner can be imported from the refiner module directly.

services/ui_backend_service/data/cache/store.py
```diff
@@ -126,65 +100,21 @@ async def get_run_parameters(self, flow_name, run_number):
...
-                run_id_value,
-                "_parameters",
-                r"\_%",
-                "name",  # exclude the 'name' parameter as this always exists, and contains the FlowName
-                "script_name"  # exclude the internally used 'script_name' parameter.
-            ],
-            fetch_single=False,
-            expanded=True
-        )
+        db_response, *_ = await self._artifact_table.get_run_parameter_artifacts(flow_name, run_number)
```
saikonen, line 110: this basically uses find_records under the hood, which accepts a postprocessing callable as a parameter. could be further simplified by passing a refiner in I guess.. but might convolute the issue further. Much cleaner this way than previously anyway :)

services/ui_backend_service/api/task.py
msavela: Good point

services/ui_backend_service/data/cache/store.py
msavela: There was a single exception catching for GetParametersFailed, which can be replaced with more generic exception catching.

services/ui_backend_service/data/cache/store.py
msavela: Very valid comment

services/ui_backend_service/data/db/tables/artifact.py
```diff
@@ -15,3 +16,48 @@ class AsyncArtifactTablePostgres(AsyncPostgresTable):
     trigger_keys = MetadataArtifactTable.trigger_keys
     select_columns = keys
     _command = MetadataArtifactTable._command
+
+    async def get_artifact_locations_for_run_ids(self, run_ids=[]):
+        # do not touch DB if no run_ids were given.
+        if len(run_ids) == 0:
```
saikonen, line 22: small flub in original code: the function actually deals with run_numbers so the param name should reflect this. run_ids will not work for the fetching. rename run_ids to run_numbers :)

saikonen (approved):

saikonen (approved):

PR #144, opened 2021-03-26
feat: data adapter enhancements for ui_service

saikonen:
Data adapter enhancements originally introduced in the UI development branch. Merge these changes only after #145 has been merged, and metadata_service has integration tests to cover basic functionality.

add shared utils module for services:
  - exposes configuration env vars as constants
  - api response wrappers
  - error handler wrappers for api
  - logger configuration
  - Database connection configuration
 
Postgres data adapter enhancements:
- update Dockerfile to include shared data module
- accept DBConfiguration in adapter initialization.
- change singleton pattern so base classes can be inherited separately.
- add logging support
- db query improvements (filters, pagination)
- Table Trigger setup support.
- Configurable table names via environment variables.

Todo:

- [ ] rebase from master after #146 has been introduced, in order to see if tests still pass with these changes.
- [ ] after merging #146  clean up pytest integration test teardown with newly introduced `execute_sql`
- [ ] add `services.utils` and `services.data` as part of the Tox test lookup after #145 is merged.


services/data/db_utils.py
```diff
@@ -29,25 +32,13 @@ def get_db_ts_epoch_str():
 
 
 def translate_run_key(v: str):
-    key = "run_id"
-    value = "'{0}'".format(v)
```
ferras, line 33: i need to double check but I believe the single quotation marks are necessary when querying the db based off run_id

msavela: I'll check this as well

msavela: Single quotation marks are not necessary due to the way we construct queries from now on at `postgres_async_db.py` https://github.com/Netflix/metaflow-service/pull/144/files/01e89fa7966259c89be9e13aca82e4cfbb93da64#diff-6bca962443914b058b03743877f9eab3f0f30804c06c086146d53d2aee6aa88aR220

This allows Psycopg handle different data types and makes it safer as well, more information here: https://www.psycopg.org/docs/usage.html#the-problem-with-the-query-parameters

msavela (approved):

ferras (approved):

services/utils/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+SERVICE_COMMIT_HASH = os.environ.get("BUILD_COMMIT_HASH", None)
+# Build time of service, if set as an environment variable.
+SERVICE_BUILD_TIMESTAMP = os.environ.get("BUILD_TIMESTAMP", None)
+
+# Setup log level based on environment variable
+log_level = os.environ.get('LOGLEVEL', 'INFO').upper()
+logging.basicConfig(level=log_level)
+
+
+async def read_body(request_content):
```
ferras, line 29: a lot of the functions in this file are present in:
https://github.com/Netflix/metaflow-service/blob/master/services/metadata_service/api/utils.py
can you remove the duplicates?

services/data/postgres_async_db.py
```diff
@@ -51,32 +64,45 @@ def __init__(self):
...
+                        pool_max=self.pool.maxsize))
+
+                break  # Break the retry loop
+            except Exception as e:
+                self.logger.exception("Exception occured")
+                if retries - i < 1:
+                    raise e
+                time.sleep(1)
+
+    async def get_table_by_name(self, table_name: str):
```
ferras, line 101: why is this method async?

services/data/db_utils.py
```diff
@@ -60,3 +51,20 @@ def get_exposed_task_id(task_id, task_name):
     if task_name is not None:
         return task_name
     return task_id
+
+
+def _get_latest_attempt_id_for_tasks(artifacts):
```
ferras, line 56: the underscore prefix in this repo was meant to signify that the function shouldn't be used outside of the class or module

PR #145, opened 2021-03-26
feat: running tests with docker-compose and configure codestyles

saikonen:
Testing and codestyle enhancements originally introduced in the UI development branch. Aim to merge this before #144 so we can write some integration tests for the metadata service.

This PR omits any pointers to `service.ui_backend` and `services.utils` as these are introduced in future PR's and including missing paths would break the test runner.

Changes: 
    * add Tox support and pycodestyle directives
    * update .gitignore for tox temp files
    * add docker-compose and dockerfile for running tests
    * add instructions to README.md for running tests
    * add placeholder test for metadata_service
    * pin service dependency versions
    * add configurable wait time and retry count to postgres server startup waiting.

### DIFF
--- a/services/migration_service/requirements.txt
+++ b/services/migration_service/requirements.txt
@@ -1,3 +1,4 @@
 aiohttp==3.6.2
 aiohttp-swagger==1.0.15
+psycopg2==2.8.6
 aiopg==1.0.0
--- a/wait-for-postgres.sh
+++ b/wait-for-postgres.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+while !</dev/tcp/${MF_METADATA_DB_HOST}/${MF_METADATA_DB_PORT}; do
+    sleep 1;
+done;
+
+$@
--- a/wait-for-postgres.sh
+++ b/wait-for-postgres.sh
@@ -1,7 +1,17 @@
 #!/bin/bash
+RETRIES=1;
+MAX_RETRIES=${POSTGRES_WAIT_MAX_RETRIES:=5};
+SLEEP_SECONDS=${POSTGRES_WAIT_SLEEP_SECONDS:=1};
 
+# Retry loop for postgres server.
 while !</dev/tcp/${MF_METADATA_DB_HOST}/${MF_METADATA_DB_PORT}; do
-    sleep 1;
+    if (($RETRIES <= $MAX_RETRIES)); then
+        echo "retry $RETRIES out of $MAX_RETRIES"
+        RETRIES=$((RETRIES+1))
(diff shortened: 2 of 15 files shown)


msavela (approved):

msavela (approved):

msavela (approved):

wait-for-postgres.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+while !</dev/tcp/${MF_METADATA_DB_HOST}/${MF_METADATA_DB_PORT}; do
+    sleep 1;
```
ferras, line 4: there probably should be an upper limit on how long we wait. In production we try 5 times by default with a 1 second sleep in between each attempt

services/migration_service/requirements.txt
```diff
@@ -1,4 +1,3 @@
-aiohttp
-aiohttp_swagger
-psycopg2
-aiopg
\ No newline at end of file
+aiohttp==3.6.2
```
ferras, line 1: is there a particular reason you pinned these dependencies?

services/migration_service/requirements.txt
```diff
@@ -1,4 +1,3 @@
-aiohttp
-aiohttp_swagger
-psycopg2
```
ferras, line 3: isn't psycopg2 still required?

wait-for-postgres.sh
saikonen: Good point. Added these as configurable env vars now (in case values need to be tuned for github test runners or other environments):

- `POSTGRES_WAIT_MAX_RETRIES` default 5
- `POSTGRES_WAIT_SLEEP_SECONDS` default 1

services/migration_service/requirements.txt
saikonen: version pinning was introduced to dependencies as a result of a particular version of aiohttp depending on a version of yarl that broke url query parameter parsing. it is debatable if the pinning is necessary for all services, as for now, only the ui backend relies heavily on parsing params.

services/migration_service/requirements.txt
saikonen: Good catch, this was dropped at some point in the UI branch and probably kept working due to all service dependencies being installed in `setup.py`. I'll reintroduce the dependency.

services/migration_service/requirements.txt
ferras: pinning dependencies in general should be avoided because it will probably cause issues down the line. Ideally we can resolve the issue with url parsing, otherwise we can pin only the dependencies we need to pin

PR #146, opened 2021-03-29
WIP: add metadata service integration tests

saikonen:
Test coverage for the metadata service component. Necessary in order to have some certainty of not breaking basic functionality with future code changes. This branch depends on changes of #145, and is a prerequisite for considering merging of #144

TODO: rebase from master after #145 is introduced, or alternatively merge this onto #145

- configure metadata service test setup and teardown
- add basic business logic test coverage to metadata service api (`CREATE, GET, LIST, HEARTBEAT`) testing both the api responses, and that data is stored in DB.

### DIFF
--- a/services/metadata_service/tests/integration_tests/utils.py
+++ b/services/metadata_service/tests/integration_tests/utils.py
@@ -71,3 +71,28 @@ async def clean_db(db: AsyncPostgresDB):
         print("DB Cleanup failed after test. This might have adverse effects on further test runs", str(error))
 
 # Test fixture helpers end
+
+
+async def assert_api_get_response(cli, path: str, status: int = 200, data: object = None):
+    """
+    Perform a GET request with the provided http cli to the provided path, assert that the status and data received are correct.
+    Expectation is that the API returns text/plain format json.
+
+    Parameters
+    ----------
+    cli : aiohttp cli
+        aiohttp test client
+    path : str
+        url path to perform GET request to
+    status : int (default 200)
+        http status code to expect from response
+    data : object
+        Any json serializable data type. will undergo json.dumps before asserting with response body.
+    """
+    response = await cli.get(path)
+
+    assert response.status == status
+    body = await response.text()
+
+    if data:
+        assert body == json.dumps(data)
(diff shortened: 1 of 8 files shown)


msavela (approved):

ferras:
will these tests be runnable using github actions? it seems like it will require a postgres db to be running

msavela:
> will these tests be runnable using github actions? it seems like it will require a postgres db to be running

Eventually yes and we already have this workflow at `ui` branch: https://github.com/Netflix/metaflow-service/blob/ui/.github/workflows/test.yml#L49

Here you can find the past runs https://github.com/Netflix/metaflow-service/actions/workflows/test.yml

For local test runs we have a separate PR https://github.com/Netflix/metaflow-service/pull/145 with `docker-compose.test.yml` which takes care postgres dependency

services/metadata_service/tests/integration_tests/utils.py
```diff
@@ -0,0 +1,251 @@
...
+    tables = [
+        db.metadata_table_postgres,
+        db.artifact_table_postgres,
+        db.task_table_postgres,
+        db.step_table_postgres,
+        db.run_table_postgres,
+        db.flow_table_postgres
+    ]
+    for table in tables:
+        await table.execute_sql(select_sql="DELETE FROM {}".format(table.table_name))
```
saikonen, line 72: @ferras should we include some kind of guard for the test suite, so this can not be run against any non-test DB? currently there are no checks in place for the DSN provided by the DBConfiguration class, rather it is implicitly expected that when running tests, the environment is configured to point to a test DB.

this could have some serious effects in case a user ends up somehow running the test suite against a production db, specifically due to the `clean_db()` teardown.

msavela: I propose that we fix the database hostname to `db_test` to avoid accidental test runs against wrong database, at least for now.

ferras: yeah i think the db name (not just the hostname) and really the rest of the configuration (password, username) should be hard coded for the test to avoid any accidents

saikonen: I opened a PR with a more thorough solution for this configuration with #167 , which should effectively namespace used environment variables by the test suite. Can be merged into this branch if acceptable.

msavela (approved):

PR #148, opened 2021-04-08
feat[dag]: Include Docstring for each DAG node

msavela:
Example DAG Node JSON object with Docstring:

```json
{
    "process_foreach": {
      "type": "linear",
      "box_next": false,
      "box_ends": null,
      "next": ["join_foreach"],
      "doc": "Multiline\n\ndocstring"
    }
}
```

### DIFF
--- a/services/ui_backend_service/data/cache/generate_dag_action.py
+++ b/services/ui_backend_service/data/cache/generate_dag_action.py
@@ -2,7 +2,6 @@ import hashlib
 import json
 from tarfile import TarFile
 
-from metaflow import FlowSpec
 from metaflow.client.cache import CacheAction
 from services.utils import get_traceback_str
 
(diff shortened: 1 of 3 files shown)


services/ui_backend_service/data/cache/generate_dag_action.py
```diff
@@ -133,7 +133,8 @@ def generate_dag(flow_id, tarball_path):
             'type': node.type,
             'box_next': node.type not in ('linear', 'join'),
             'box_ends': node.matching_join,
-            'next': node.out_funcs
+            'next': node.out_funcs,
+            'doc': node.doc
```
rsanteri, line 137: Is it possible here that node.doc is not defined or is None?

rsanteri: Meaning should we have fallback for that?

msavela: It will always default to empty string, this is handled here: https://github.com/Netflix/metaflow-service/blob/44e724600e5fdee0b76850eafe31d70fe79f993f/services/ui_backend_service/data/cache/custom_flowgraph.py#L158

PR #150, opened 2021-04-12
reliability improvements

ferras:
add healthcheck endpoint
make startup and connection attempts configurable via env variables


services/metadata_service/api/admin.py
```diff
@@ -48,6 +52,43 @@ async def ping(self, request):
...
+            await cur.execute("SELECT 1")
+            records = await cur.fetchall()
+            if len(records) > 0:
+                status["status"] = "UP"
+            else:
+                status["status"] = "DOWN"
+                status_code = 500
+
+            cur.close()
+        return web.Response(status=status_code, body=json.dumps(status))
```
saikonen, line 89: I recall having issues with writing tests for routes using the basic web.Response, as this does not set the correct content-type headers automatically. The upcoming `services.utils#web_response` helper that is used overall for json routes gets around this issue.

Seems acceptable for now, especially if nothing breaks due to the mismatched content-type header. I would suggest leaving a TODO for once the utils module and tests tooling is merged in.

services/data/service_configs.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+max_connection_retires = os.environ.get("MF_SERVICE_CONNECTION_RETRIES", 3)
+connection_retry_wait_time_seconds = os.environ.get("MF_SERVICE_CONNECTION_RETRY_WAITTIME_SECONDS", 1)
+max_startup_retries = os.environ.get("MF_SERVICE_STARTUP_RETRIES", 5)
+startup_retry_wait_time_seconds = os.environ.get("MF_SERVICE_STARTUP_WAITTIME_SECONDS", 1)
```
saikonen, line 6: Consider casting these to correct types. For example setting `MF_SERVICE_CONNECTION_RETRIES=1` would presumably fail in the `postgres_async_db` due to a string being passed to `range`

PR #161, opened 2021-04-20
UI service dependency pin

msavela:
Remove dependency version pinning from ui service and drop unnecessary package.json files.


services/ui_backend_service/requirements.txt
```diff
@@ -1,7 +1,6 @@
-aiohttp==3.6.2
-aiohttp-swagger==1.0.15
-boto3==1.15.10
-pyee==8.0.1
-yarl==1.5.1
```
saikonen, line 5: is the latest yarl version functional again regarding url parameters?

services/ui_backend_service/requirements.txt
```diff
@@ -1,7 +1,6 @@
-aiohttp==3.6.2
-aiohttp-swagger==1.0.15
-boto3==1.15.10
```
saikonen, line 3: I think the reason for pinning the boto3 version had something to do with the dependencies introduced with the cache client, will check what this was about.

saikonen (approved):

services/ui_backend_service/requirements.txt
msavela: `aiohttp` is now working as expected without pinning down `yarl` version.

PR #181, opened 2021-05-24
UI cache error handling

msavela:
This PR introduces following changes:

- Prevent `CacheServerUnreachable` fatal error.
- Cache `max_size` is now considered as a soft limit, GC will eventually delete exceeding data after graceful period.
- Keep track of pending cache actions so that `CacheClient` and `CacheFuture` can provide accurate responses.

### DIFF
--- a/services/ui_backend_service/data/cache/client/cache_server.py
+++ b/services/ui_backend_service/data/cache/client/cache_server.py
@@ -131,10 +131,14 @@ class Worker(object):
         self.prio = request['priority']
         self.filestore = filestore
         self.proc = None
-        self.tempdir = self.filestore.open_tempdir(request['idempotency_token'],
-                                                   request['action'],
-                                                   request['stream_key'])
-        if self.tempdir is None:
+
+        try:
+            self.tempdir = self.filestore.open_tempdir(
+                request['idempotency_token'],
+                request['action'],
+                request['stream_key'])
+        except Exception:
+            self.tempdir = None
             self.echo("Store couldn't create a temp directory. "
                       "WORKER NOT STARTED.")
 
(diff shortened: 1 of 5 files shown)


services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -234,26 +262,19 @@ def validate_actions(self, actions):
...
     def schedule(self):
-
         def queued_request(queue):
             while queue:
                 yield queue.popleft()
 
         for request in chain(queued_request(self.hi_prio_requests),
                              queued_request(self.lo_prio_requests)):
 
-            try:
```
msavela, line 245: This PR will allow cache size to exceed `max_size` and rely on GC to cleanup old data once graceful period is over. 
I'm not sure which approach would be the better one.

services/ui_backend_service/data/cache/client/cache_store.py
```diff
@@ -170,42 +169,43 @@ def ensure_path(self, path):
...
+                # make sure that the new symlink points at a valid (empty!)
+                # file by creating a dummy destination file
+                self.ensure_path(src)
+                open_res = self.safe_fileop(open, dst, 'w')
+                if open_res:
+                    _, f = open_res
+                    f.close()
+                    try:
+                        os.symlink(dst, src)
+                    except Exception as ex:
```
saikonen, line 196: originally caught only OSError, as Exception does not have `errno`. Not sure if symlink can raise anything except OSError though, so this might not be an issue

services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -234,26 +262,19 @@ def validate_actions(self, actions):
...
-                    # NOTE: Cleanup this logic.
-                    worker.start()
-                    return worker
-            except CacheFullException:
-                echo("Cache storage is full, worker failed to start!")
-                return None
+            worker = Worker(request, self.filestore)
+            if worker.tempdir:
+                worker.start()
+                return worker
```
saikonen, line 275: will the `start()` always succeed? the `new_worker()` in `loop()` also has some optimistic assumptions about workers.

services/ui_backend_service/data/cache/client/cache_store.py
msavela: Good catch, this is most likely due to older implementation which originated from the cache fork.

services/ui_backend_service/data/cache/client/cache_server.py
msavela: Additional error handling now in place.

saikonen (approved):

PR #184, opened 2021-05-28
Service cleanup

romain-intel:


services/metadata_service/api/admin.py
```diff
@@ -96,11 +96,11 @@ async def healthcheck(self, request):
     async def get_authorization_token(self, request):
         """
         ---
-        description: this is used exclusively for sandbox auth
+        description: This endpoint is used exclusively for sandbox auth
         tags:
         - Auth
         produces:
-        - text/plain
+        - application/json
```
savingoyal, line 103: Have you verified that this change wouldn't affect the sandbox? We have a similar issue where the `heartbeat` endpoint returns an actual JSON but is tagged as `text/plain` which has caused issues in the metaflow codebase before.

services/metadata_service/api/artifact.py
```diff
@@ -78,10 +77,12 @@ async def get_artifact(self, request):
           required: true
           type: "string"
         produces:
-        - text/plain
+        - application/json
```
savingoyal, line 80: Same as the comment above.

savingoyal (changes requested):

services/metadata_service/api/artifact.py
```diff
@@ -306,16 +312,18 @@ async def create_artifacts(self, request):
...
-            task_id, task_name = await self._db.get_task_ids(flow_name, run_number,
-                                                             step_name, task_id)
-        except Exception:
-            return web.Response(status=400, body=json.dumps(
-                {"message": "need to register run_id and task_id first"}))
+        run = await self._db.get_run_ids(flow_name, run_number)
+        task = await self._db.get_task_ids(flow_name, run_number,
+                                            step_name, task_id)
+        if run.response_code != 200 or task.response_code != 200:
+            return DBResponse(400, {"message": "need to register run_id and task_id first"})
```
saikonen, line 322: A more generic thought and not a blocker for merging the cleanup, but is there a reason why this kind of integrity check is not covered by a foreign key constraint in the table?

also affects the metadata create handler

services/metadata_service/api/utils.py
```diff
@@ -53,10 +54,20 @@ def handle_exceptions(func):
...
-    async def wrapper(*args, **kwargs):
+    async def wrapper(self, request):
         try:
-            return await func(*args, **kwargs)
+            return await func(self, request)
         except Exception as err:
-            return http_500(str(err))
+            # either use provided traceback from subprocess, or generate trace from current process
+            err_trace = getattr(err, 'traceback_str', None) or get_traceback_str()
+            print(err_trace)
```
saikonen, line 63: Necessary print or leftover from some testing? If this is for logging purposes, consider using the logger from `services.utils` instead. 

If this is for generic logging over the process, it should probably be set as a logging handler with `loop.set_exception_handler(some_error_handling_function)` in the server setup phase instead. You can see https://github.com/Netflix/metaflow-service/blob/ui/services/ui_backend_service/ui_server.py#L98 for an example

services/utils/__init__.py
```diff
@@ -18,15 +18,6 @@
...
 
-async def read_body(request_content):
-    byte_array = bytearray()
-    while not request_content.at_eof():
-        data = await request_content.read(4)
-        byte_array.extend(data)
-
-    return json.loads(byte_array.decode("utf-8"))
-
-
```
saikonen, line 29: Were there some issues encountered with using `await response.json()` or why was this previously necessary? Concerned about some possible edge cases that motivated the previous implementation

services/metadata_service/api/admin.py
```diff
@@ -139,5 +139,5 @@ async def get_authorization_token(self, request):
 
             return web.Response(status=200, body=json.dumps(credentials))
         except Exception as ex:
-            body = {"err_msg": str(ex), "traceback": get_traceback_str()}
+            body = {"message": str(ex), "traceback": get_traceback_str()}
```
saikonen, line 142: is this the expected body for the authorization route? the http_500 helper only has a `detail` field for the error message. Consider keeping it consistent with the helper output if possible?

services/metadata_service/api/utils.py
```diff
@@ -53,10 +54,20 @@ def handle_exceptions(func):
...
+            # We log the request that caused this for debugging information
+            try:
+                body = await request.text()
+            except:
+                body = '<no body>'
+            print("Error caused when %s %s with query %s and body %s" %
```
saikonen, line 69: same deal here, could be using the logger instead to adhere to configured log levels.

services/metadata_service/api/step.py
```diff
@@ -144,12 +144,16 @@ async def create_step(self, request):
...
-        body = await read_body(request.content)
+        body = await request.json()
         user = body.get("user_name", "")
         tags = body.get("tags")
         system_tags = body.get("system_tags")
 
-        run_number, run_id = await self._db.get_run_ids(flow_name, run_number)
+        run = await self._db.get_run_ids(flow_name, run_number)
+        if run.response_code != 200:
+            return DBResponse(400, {"message": "need to register run_id first"})
```
saikonen, line 154: Steps have a foreign key for runs at least in the V3 schema, so consider handling error code 409 specifically for this instead? 

Is it necessary to rewrite the error messages on the response handler level, or could this be handled at a higher level in `aiopg_exception_handling` instead?

services/metadata_service/api/task.py
```diff
@@ -181,7 +181,11 @@ async def create_task(self, request):
             return web.Response(status=400, body=json.dumps(
                 {"message": "provided task_name may not be a numeric"}))
 
-        run_number, run_id = await self._db.get_run_ids(flow_id, run_number)
+        run = await self._db.get_run_ids(flow_id, run_number)
+        if run.response_code != 200:
+            return DBResponse(400, {"message": "need to register run_id and task_id first"})
```
saikonen, line 186: same comment as for steps. Tasks in the V3 schema also have a foreign key covering this, so consider handling the specific error instead.

services/utils/__init__.py
romain-intel: I will check with Ferras for the why of the previous implementation but I felt it would be better to use the standard implementation instead of doing our own.

services/metadata_service/api/admin.py
romain-intel: I am trying to make them all consistent. Still working on it. After I pushed this, I realized I missed af ew things.

services/metadata_service/api/utils.py
romain-intel: Yep, will change to a logger here and elsewhere. I will try to get the proper one (I couldn't find it initially).

services/metadata_service/api/step.py
romain-intel: I am reworking this and yes, plan to raise an exception all the way through. The issue before though was that it would be silent.

services/metadata_service/api/utils.py
romain-intel: I will use the logger. I did want to log things to aid in debugging after.

services/metadata_service/api/artifact.py
romain-intel: It is covered I think; the issue is that it used to be silent (ie: the db would raise the error but it could silently make it through.

services/metadata_service/api/admin.py
romain-intel: I am going to rework things and push a PR to make it so that everything works on application/json. I will also check with Ferras if there was a reason for the string only implementation.

romain-intel:
Thanks for all the comments. I am going to keep working on this. After I pushed it out, I realized it was still not as clean as I wanted it to be and that I would need something on the Metaflow side to talk properly to an actual REST service.

PR #185, opened 2021-05-28
Add a 'tags' endpoint to support runtime tagging

romain-intel:
This needs #184 to function properly.

### DIFF
--- a/services/metadata_service/api/tag.py
+++ b/services/metadata_service/api/tag.py
@@ -4,7 +4,6 @@ from services.data.postgres_async_db import AsyncPostgresDB
 from services.metadata_service.api.utils import format_response, \
     handle_exceptions
 import json
-from aiohttp import web
 
 import asyncio
 
@@ -98,14 +97,14 @@ class TagApi(object):
                     raise ValueError("invalid artifact specification: %s" % o['id'])
                 obj_filter = table.get_filter_dict(*pathspec)
             except ValueError as e:
-                return web.Response(status=400, body=json.dumps(
+                return DBResponse(response_code=400, body=json.dumps(
                     {"message": "invalid input: %s" % str(e)}))
 
             # Now we can get the object
             obj = await table.get_records(
                 filter_dict=obj_filter, fetch_single=True, expanded=True)
             if obj.response_code != 200:
-                return web.Response(status=obj.response_code, body=json.dumps(
+                return DBResponse(response_code=obj.response_code, body=json.dumps(
                     {"message": "could not get object %s: %s" % (o['id'], obj.body)}))
 
             # At this point do some checks and update the tags
@@ -115,7 +114,7 @@ class TagApi(object):
                 # This is the only error we fail hard on; adding a tag that is
                 # in system tag
                 if o['tag'] in obj['system_tags']:
-                    return web.Response(status=405, body=json.dumps(
+                    return DBResponse(response_code=405, body=json.dumps(
                         {"message": "tag %s is already a system tag and can't be added to %s"
                         % (o['tag'], o['id'])}))
                 if o['tag'] not in obj['tags']:
@@ -126,15 +125,15 @@ class TagApi(object):
                     modified = True
                     obj['tags'].remove(o['tag'])
             else:
-                return web.Response(status=400, body=json.dumps(
+                return DBResponse(response_code=400, body=json.dumps(
                     {"message": "invalid tag operation %s" % o['operation']}))
             if modified:
                 # We save the value back
                 result = await table.update_row(filter_dict=obj_filter, update_dict={
                     'tags': "'%s'" % json.dumps(obj['tags'])})
                 if result.response_code != 200:
-                    return web.Response(status=result.response_code, body=json.dumps(
+                    return DBResponse(response_code=result.response_code, body=json.dumps(
                         {"message": "error updating tags for %s: %s" % (o['id'], result.body)}))
             results.append(obj)
 
-        return DBResponse(response_code=200, body=results)
+        return DBResponse(response_code=200, body=json.dumps(results))
--- a/services/metadata_service/api/tag.py
+++ b/services/metadata_service/api/tag.py
@@ -111,26 +111,20 @@ class TagApi(object):
             obj = obj.body
             modified = False
             if o['operation'] == 'add':
-                # This is the only error we fail hard on; adding a tag that is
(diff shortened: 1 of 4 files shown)


services/metadata_service/api/tag.py
```diff
@@ -0,0 +1,139 @@
...
+                if o['tag'] in obj['tags']:
+                    modified = True
+                    obj['tags'].remove(o['tag'])
+            else:
+                return DBResponse(response_code=400, body=json.dumps(
+                    {"message": "invalid tag operation %s" % o['operation']}))
+            if modified:
+                # We save the value back
+                result = await table.update_row(filter_dict=obj_filter, update_dict={
+                    'tags': "'%s'" % json.dumps(obj['tags'])})
```
saikonen, line 133: Worth checking: Previously update_row was only used for heartbeat updating, but using it this way might not be reliable/secure, as tags contain user inputs. Preferred way for `psycopg` is to use the `cur.execute(sql_template, values)` which should take care of correctly escaping the values.

Might be worth refactoring the update_row function to use the safe execute syntax.
see https://www.psycopg.org/docs/usage.html#passing-parameters-to-sql-queries for usage.

services/metadata_service/api/tag.py
```diff
@@ -0,0 +1,139 @@
...
+from services.data.postgres_async_db import AsyncPostgresDB
+from services.metadata_service.api.utils import format_response, \
+    handle_exceptions
+import json
+
+import asyncio
+
+
+class TagApi(object):
+    lock = asyncio.Lock()
```
saikonen, line 12: seems to be unused.

services/metadata_service/api/tag.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, app):
+        app.router.add_route(
+            "POST",
+            "/tags",
+            self.update_tags,
+        )
```
saikonen, line 19: Is there a future-proofing reason why we want a single route that can update multiple resource tags in bulk, instead of the more RESTful approach of accepting a whitelisted PUT/PATCH to the specific resource routes? f.ex. updating a tasks tags could be done with a 
```PATCH /flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}```

Or if it is necessary to have a specific tag update route (even though tags are part of the resource in question and not a separate entry)
```PATCH /flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/tags```

File: services/metadata_service/api/tag.py [hunk 4]
```diff
@@ -0,0 +1,139 @@
...
+                    return DBResponse(response_code=405, body=json.dumps(
+                        {"message": "tag %s is already a system tag and can't be added to %s"
+                        % (o['tag'], o['id'])}))
+                if o['tag'] not in obj['tags']:
+                    modified = True
+                    obj['tags'].append(o['tag'])
+            elif o['operation'] == 'remove':
+                if o['tag'] in obj['tags']:
+                    modified = True
+                    obj['tags'].remove(o['tag'])
```
saikonen, line 126: Noticed that the Metaflow client and service do not guard against duplicate tags, so `--tag test --tag test` would add the test tag twice, due to tags being a simple array. `remove` will only get rid of one occurrence of a tag. Should this be the way the update works?

alternatively get rid of all occurences of a tag:
```
obj['tags'] = [tag for tag in obj['tags] if tag != o['tag']]
```

services/metadata_service/api/tag.py
```diff
@@ -0,0 +1,139 @@
...
+                description: successful operation. Return newly registered task
+            "404":
+                description: not found
+            "500":
+                description: internal server error
+        """
+        body = await request.json()
+        results = []
+        for o in body:
+            try:
```
saikonen, line 84: The multiple updates are not done inside a transaction, so there is a slight peculiarity with the way things might work in case of errors. For example

1. flow/run/step tag updates succeed
2. for whatever reason task tag update fails
3. API responds with an error, even though some records were updated, and some updates were never reached.

services/metadata_service/api/tag.py
saikonen: On further thought this approach would lead to an immense number of requests in order to update tags for all tasks of a run. Feel free to ignore :)

services/metadata_service/api/tag.py
```diff
@@ -0,0 +1,139 @@
...
+            # At this point do some checks and update the tags
+            obj = obj.body
+            modified = False
+            if o['operation'] == 'add':
+                # This is the only error we fail hard on; adding a tag that is
+                # in system tag
+                if o['tag'] in obj['system_tags']:
+                    return DBResponse(response_code=405, body=json.dumps(
+                        {"message": "tag %s is already a system tag and can't be added to %s"
+                        % (o['tag'], o['id'])}))
```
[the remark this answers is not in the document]
romain-intel: That's an excellent point and a bug most likely. This patch was done to allow modification of tags after the fact and the idea is to make sure that system tags stay unique and not affected. They are kept separate but we were trying to make sure that system tags were not overwritten in a way although I will admit that this is maybe not very useful. Let me circle back on that.

services/metadata_service/api/tag.py
romain-intel: Good point. I will update the update_row function to use a template. That seems the better way.

services/metadata_service/api/tag.py
romain-intel: I copied from other APIs but let me check if there was a purpose behind it.

services/metadata_service/api/tag.py
romain-intel: Ya, I think the idea was to allow bulk updates and since the tags can be added at lots of level it may make a lot of endpoints/calls for essentially the same thing.

services/metadata_service/api/tag.py
romain-intel: You are right again on the creation side of things; the update will guard against duplicates but not the add (I will double check on that side; I think we should just convert it to a set and set that and also check against system tags).

services/metadata_service/api/tag.py
romain-intel: Yes, this was a replica of how things were happening inside (in our internal version) and was a conscious choice but maybe something worth reconsidering. @ferras may have some more ideas on that.

PR #187, opened 2021-06-10
CacheAsyncClient increase pipe buffer

msavela:
- Increase CacheAsyncClient pipe buffer from 64KB to 1024KB
- Prevent passing of individual artifact keys from subprocess to ensure buffer is not exceeded

This PR is a precautionary measure, actual issue which is the cache limit should not be configured to exceed available disk space. Configured with env variables:
- `CACHE_ARTIFACT_STORAGE_LIMIT`
- `CACHE_DAG_STORAGE_LIMIT`

### DIFF
--- a/services/ui_backend_service/data/cache/client/cache_server.py
+++ b/services/ui_backend_service/data/cache/client/cache_server.py
@@ -208,7 +208,7 @@ class Worker(object):
 
     def _worker_details(self):
         return {
-            'keys': self.request['keys'],
+            'keys': len(self.request['keys']),
             'stream_key': self.request['stream_key'],
             'idempotency_token': self.request["idempotency_token"],
         }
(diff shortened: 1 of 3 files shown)


services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -208,7 +208,7 @@ def kill(self):
 
     def _worker_details(self):
         return {
-            'keys': self.request['keys'],
+            'keys': len(self.request['keys']),
```
saikonen, line 211: is this only used for logging purposes, so returning the length instead of the keys is not an issue?

saikonen: to answer my own comment, yes, seems to be the case.

saikonen (approved):

services/ui_backend_service/data/cache/client/cache_server.py
msavela: This is only for logging/informational purposes. This way we can ensure fairly fixed amount of data transported through the pipe instead of highly variable amount of data. This ended up exceeding the buffer in some cases when there were hundreds of artifacts to fetch.

PR #189, opened 2021-06-11
fix: tags autocomplete performance enhancement

saikonen:
Changes the tags autocomplete routes to use in-memory cached tags and list operations for filtering. Backend fetches fresh tags periodically. Initial benchmarks show that this performs immensely better than the query-only approach (~300ms vs 5sec latencies)

### DIFF
--- a/services/ui_backend_service/api/autocomplete.py
+++ b/services/ui_backend_service/api/autocomplete.py
@@ -14,7 +14,7 @@ class AutoCompleteApi(object):
         self.db = db
         # Cached resources
         # Cache tags so we don't have to request DB everytime
-        self.tags = None
+        self.tags = []
         app.router.add_route("GET", "/tags/autocomplete", self.get_tags)
         # Non-cached resources
         app.router.add_route("GET", "/flows/autocomplete", self.get_flows)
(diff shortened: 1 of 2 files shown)


services/ui_backend_service/api/autocomplete.py
```diff
@@ -1,21 +1,43 @@
...
+        Async task that fill tags cache every 5minutes. Database query might take a while
+        so its better to cache the result.
+        '''
+        while True:
+            # Get all tags that are mentioned in runs table
+            res, _ = await self.db.run_table_postgres.get_tags()
+            if res.response_code == 200:
+                self.tags = sorted(res.body)
+            size = sys.getsizeof(self.tags) // 1024 // 1024
+            print("SIZE OF TAGS IN MEM: {} Mb".format(size), flush=True)
```
msavela, line 38: Maybe use logger here at debug or info level

msavela (approved):

services/ui_backend_service/api/autocomplete.py
saikonen: this is just some leftovers from the WIP. Would it be beneficial to actually log out the object size after fetching though? There's something odd going on with sys.getsizeof as well, seems to work locally, but reports 0 mb on the deployed version. for whatever reason.

saikonen: changed this to use a logger. printing out the memory used, and the number of tags stored with an `INFO` log level

msavela (approved):

PR #191, opened 2021-06-15
fix[notify]: Recover from connection failures to db.

msavela:
This PR fixes a problem where real time events from the database via `pg_notify` failed to recover from connection failures. Try to reconnect every 1s once cursor is detected as closed.

### DIFF
--- a/services/ui_backend_service/api/notify.py
+++ b/services/ui_backend_service/api/notify.py
@@ -1,7 +1,6 @@
 import json
-import time
 import asyncio
-from typing import Dict, List
+from typing import Dict
 from services.utils import logging
 from services.data.postgres_async_db import (
     FLOW_TABLE_NAME, RUN_TABLE_NAME, STEP_TABLE_NAME, TASK_TABLE_NAME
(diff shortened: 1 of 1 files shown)


saikonen (approved):

services/ui_backend_service/api/notify.py
```diff
@@ -34,16 +33,23 @@ def __init__(self, app, db, event_emitter=None):
...
 
     async def _init(self, pool):
-        async with pool.acquire() as conn:
-            await asyncio.gather(
-                self.listen(conn),
-                self.ping(conn)
-            )
+        while True:
+            try:
+                async with pool.acquire() as conn:
```
saikonen, line 38: can this cause some unexpected effects? the `while` / `async with` is in a different order compared to the previous implementation

msavela: I'm not completely sure what you mean, there wasn't a `while` loop in the previous implementation. Once coroutines(listen/ping) are done execution leaves the context manager and previous connection is closed before new connection is acquired from the pool.

PR #192, opened 2021-06-18
fix[cache]: Cache worker recovery by restart.

msavela:
Cache worker drain timeout error indicates unrecoverable critical issue, essentially the cache functionality remains broken after the first `asyncio.TimeoutError` (drain lock) occures.

`CacheAsyncClient` can now request restart for itself and `CacheStore` will monitor these requests periodically (every 5s) so that normal operation can be resumed.

### DIFF
--- a/services/ui_backend_service/data/cache/client/cache_async_client.py
+++ b/services/ui_backend_service/data/cache/client/cache_async_client.py
@@ -15,6 +15,7 @@ HEARTBEAT_FREQUENCY = 1
 
 class CacheAsyncClient(CacheClient):
     _drain_lock = asyncio.Lock()
+    _restart_requested = False
 
     async def start_server(self, cmdline, env):
         self.logger = logging.getLogger("CacheAsyncClient:{root}".format(root=self._root))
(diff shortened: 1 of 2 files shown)


services/ui_backend_service/data/cache/client/cache_async_client.py
```diff
@@ -15,6 +15,7 @@
 
 class CacheAsyncClient(CacheClient):
     _drain_lock = asyncio.Lock()
+    _restart_requested = False
```
saikonen, line 18: as this lives in the asyncClient instance, does anything set it back to false after a successful restart? To my understanding the cacheclient instance does not get reinitialized on a restart?

msavela: Actually it does get reinitialized on restart. This is the relevant line: https://github.com/Netflix/metaflow-service/blob/dcbd511267a45226d4e29d67f1ae2667e58e1147/services/ui_backend_service/data/cache/store.py#L115

saikonen: so it seems, disregard the comment then ;)

saikonen (approved):

PR #195, opened 2021-06-23
fix[cache]: Get artifact error report for individual items.

msavela:
Cache functionality changes:
- More accurate error results
- Each artifact error result is now returned individually instead of one global error causing whole request to fail
- Added support for streaming errors in real-time
- Added support for cache invalidation via `?invalidate=true` query parameter where results are cacheable
- Timeline search result now includes individual errors instead one global failure with error message for each task

### DIFF
--- a/services/ui_backend_service/api/utils.py
+++ b/services/ui_backend_service/api/utils.py
@@ -420,7 +420,8 @@ async def find_records(request: web.BaseRequest, async_table=None, initial_condi
     conditions = initial_conditions + builtin_conditions + custom_conditions
     values = initial_values + builtin_vals + custom_vals
     ordering = (initial_order or []) + (order or [])
-    benchmark = request.query.get("benchmark", False) in ['True', 'true', '1', "t"]
+    benchmark = query_param_enabled(request, "benchmark")
+    invalidate_cache = query_param_enabled(request, "invalidate")
 
     results, pagination, benchmark_result = await async_table.find_records(
         conditions=conditions, values=values, limit=limit, offset=offset,
@@ -428,6 +429,7 @@ async def find_records(request: web.BaseRequest, async_table=None, initial_condi
         fetch_single=fetch_single, enable_joins=enable_joins,
         expanded=True,
         postprocess=postprocess,
+        invalidate_cache=invalidate_cache,
         benchmark=benchmark,
         overwrite_select_from=overwrite_select_from
     )
@@ -441,6 +443,11 @@ async def find_records(request: web.BaseRequest, async_table=None, initial_condi
         res["benchmark_result"] = benchmark_result
 
     return web_response(status, res)
+
+
+def query_param_enabled(request: web.BaseRequest, name: str) -> bool:
+    """Parse boolean query parameter and return enabled status"""
+    return request.query.get(name, False) in ['True', 'true', '1', "t"]
 
 
 class TTLQueue:
--- a/services/ui_backend_service/data/cache/client/cache_client.py
+++ b/services/ui_backend_service/data/cache/client/cache_client.py
@@ -47,7 +47,7 @@ class CacheFuture(object):
         return bool(self.stream_key)
 
     def wait(self, timeout=FOREVER):
-        return self.client.wait(lambda: True if self.is_ready() else None,
+        return self.client.wait(lambda: None if self.has_pending_request() else True,
                                 timeout)
 
     def get(self):
@@ -183,7 +183,8 @@ class CacheClient(object):
             msg, keys, stream_key, disposable_keys =\
                 cls.format_request(*args, **kwargs)
             future = CacheFuture(keys, stream_key, self, cls, self._root)
-            if future.key_paths_ready():
+            invalidate_cache = msg.get('invalidate_cache', False)
+            if future.key_paths_ready() and not invalidate_cache:
                 # cache hit
                 req = None
             else:
--- a/services/ui_backend_service/data/cache/client/cache_client.py
+++ b/services/ui_backend_service/data/cache/client/cache_client.py
@@ -55,10 +55,12 @@ class CacheFuture(object):
             with open(path, 'rb') as f:
                 return f.read()
 
+        _safe_key_paths = {key: path for key, path in self.key_paths.items() if is_safely_readable(path)}
         if self.key_objs is None and self.is_ready():
             self.key_objs = {key: _read(path)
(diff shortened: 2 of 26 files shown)


services/ui_backend_service/data/cache/get_artifacts_action.py
```diff
@@ -100,41 +100,44 @@ def stream_error(err, id, traceback=None):
...
+                    except MetaflowS3AccessDenied as ex:
+                        results[artifact_key] = json.dumps([False, 's3-access-denied', location])
+                    except MetaflowS3NotFound as ex:
+                        results[artifact_key] = json.dumps([False, 's3-not-found', location])
+                    except MetaflowS3URLException as ex:
+                        results[artifact_key] = json.dumps([False, 's3-bad-url', location])
+                    except MetaflowS3CredentialsMissing as ex:
+                        results[artifact_key] = json.dumps([False, 's3-missing-credentials', location])
+                    except MetaflowS3Exception as ex:
+                        results[artifact_key] = json.dumps([False, 's3-generic-error', get_traceback_str()])
```
saikonen, line 130: Caching all the error cases as well is very much undesirable in the long run as I think there is currently no way to invalidate a cached result?
I see no harm for this in the short term though, as the cache storage is transient due to it being part of the deployment, and if an admin is fixing S3 access issues on a deployed service, they will most likely redeploy, thus flushing the cache.

services/ui_backend_service/data/refiner/refinery.py
```diff
@@ -84,3 +83,27 @@ async def _postprocess(self, response: DBResponse):
 
         return DBResponse(response_code=response.response_code,
                           body=body)
+
+
+def unpack_processed_value(value) -> Tuple[bool, str, int]:
+    return (list(value) + [None] * 3)[:3]
```
saikonen, line 89: maybe a comment or something, what is the value this takes? use case is quite unclear on first read.

services/ui_backend_service/data/refiner/artifact_refiner.py
```diff
@@ -28,48 +18,43 @@ def __init__(self, cache):
...
-            data = {}
-            fetch_error = format_error_body(getattr(ex, "id", None), str(ex))
+        errors = {}
+
+        def _event_stream(event):
+            if event.get("type") == "error" and event.get("key"):
+                loc = artifact_location_from_key(event["key"])
+                errors[loc] = event
+
+        responses = await self.fetch_data(locations, event_stream=_event_stream)
```
saikonen, line 33: is there currently any event stream that can come out of the cache action? or is this a remnant from some testing of streaming errors?

services/ui_backend_service/data/refiner/refinery.py
```diff
@@ -84,3 +83,41 @@ async def _postprocess(self, response: DBResponse):
 
         return DBResponse(response_code=response.response_code,
                           body=body)
+
+
+def unpack_processed_value(value) -> Tuple[bool, str, Any]:
```
saikonen, line 88: is the use of `Any` justified here? from the failure examples it looks like the last two returns should both be `Optional(str)` ?

msavela: `Tuple[bool, Optional[Any], Optional[Any]]` is preferred here since `value` and `details` can be anything.

services/ui_backend_service/data/refiner/artifact_refiner.py
msavela: As of now, there's no error streaming enabled for any of the actions.

I recommend leaving this as is, since the streaming errors are very useful for future, let's say we fetch all the timeline tasks via websocket request, we could essentially broadcast errors in realtime instead of reporting them after each batch is completed.

services/ui_backend_service/api/utils.py
```diff
@@ -316,6 +318,11 @@ async def find_records(request: web.BaseRequest, async_table=None, initial_condi
     return web_response(status, res)
 
 
+def query_param_enabled(request: web.BaseRequest, name: str) -> bool:
+    """Parse boolean query parameter and return enabled status"""
+    return request.query.get(name, False) in ['True', 'true', '1', "t"]
+
+
```
saikonen, line 325: excellent cleanup. A lot of routes use this helper now so covering the intended behavior with an unit test might be benefitial, just to check that it returns a correct boolean even with unexpected parameters

services/ui_backend_service/data/cache/client/cache_client.py
```diff
@@ -181,7 +183,8 @@ def _call(*args, **kwargs):
             msg, keys, stream_key, disposable_keys =\
                 cls.format_request(*args, **kwargs)
             future = CacheFuture(keys, stream_key, self, cls, self._root)
-            if future.key_paths_ready():
+            invalidate_cache = msg.get('invalidate_cache', False)
+            if future.key_paths_ready() and not invalidate_cache:
```
saikonen, line 187: so I'm assuming that the reason the `invalidate_cache` has to be a part of the msg instead of its own return, is so that it can be passed in to the CacheAction `execute()`? Alternatively we could introduce this as part of the CacheAction API, and require the `format_response` to always return a boolean for cache retention, and `execute` to also require it as a parameter.

Current approach works fine as well, just wondering which one makes it more clear. The base CacheAction class gives no hints to the existence of such a feature as is, which would leave new developers unaware of the feature unless they dig deeper under the hood.

services/ui_backend_service/data/cache/generate_dag_action.py
```diff
@@ -78,6 +79,7 @@ def execute(cls,
                 keys=None,
                 existing_keys={},
                 stream_output=None,
+                invalidate_cache=False,
```
saikonen, line 82: api route accepts, and passes the invalidation boolean to the action. Action does not use it for anything though?

msavela: This specific action does not have any logic for existing keys(unlike other actions), it will always refetch given location(s). Normally this action will not be executed at all if key exists on disk. This line https://github.com/Netflix/metaflow-service/blob/fix/get-artifact-error-report/services/ui_backend_service/data/cache/client/cache_client.py#L186 is the one that matters in this case.

saikonen: ah I see, this action was using the stream_error and not caching errors to begin with.

saikonen (approved):

PR #196, opened 2021-06-28
feat: functional filters for websocket subscriptions

saikonen:
Removes the reliance on a database roundtrip for web socket subscriptions when determining if a record should be broadcast or not. Perform the filter checking in-process instead, with pure filter functions

Quick test on local machine with 10_000 random records: With individual calls the DB filter took 10seconds, whereas the pure function filters finished in 0.09sec


services/ui_backend_service/api/utils.py
```diff
@@ -213,6 +215,130 @@ def builtin_conditions_query_dict(query: MultiDict):
...
+    return lambda item: _filter(item[key], term) if key in item else False
+
+
+# NOTE: keep these as simple comparisons,
+# any kind of value decoding should be done outside the lambdas instead
+# to promote reusability.
+array_filter_ops = {
+    "eq": (lambda item, term: str(item) == term),
+    "ne": (lambda item, term: str(item) != term),
+    "lt": (lambda item, term: int(item) < int(term)),
```
msavela, line 230: Any concerns when comparing string values here? Thing that comes to mind first would be run_id/run_number and task_id/task_number

msavela: And what happens if we want to cast non-numerical values to integer, are we catching errors in safe manner,

saikonen: good point on the type casting gotcha. the 'term' in the filters will originate from url params, so this will be a string by default to my knowledge, so comparing values as strings by default should not be a major issue here. I'll add a catch for the possible type-cast failures

msavela (approved):

PR #199, opened 2021-07-07
Make sure aiohttp web response body is a binary

oavdeev:
If you look at [aiohttp docs](https://docs.aiohttp.org/en/stable/web_reference.html#response), body is supposed to be a binary, not string. It mostly works if you pass a string, however some auxiliary things break -- I've ran into some issues because of this while trying to use `aiohttp-devel`.

### DIFF
--- a/services/data/db_utils.py
+++ b/services/data/db_utils.py
@@ -14,17 +14,17 @@ def aiopg_exception_handling(exception):
     body = {"err_msg": err_msg}
     if isinstance(exception, psycopg2.IntegrityError):
         if "duplicate key" in err_msg:
-            return DBResponse(response_code=409, body=json.dumps(body))
+            return DBResponse(response_code=409, body=body)
         elif "foreign key" in err_msg:
-            return DBResponse(response_code=404, body=json.dumps(body))
+            return DBResponse(response_code=404, body=body)
         else:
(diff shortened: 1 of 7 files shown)


services/data/db_utils.py
```diff
@@ -14,17 +14,17 @@ def aiopg_exception_handling(exception):
...
         else:
-            return DBResponse(response_code=500, body=json.dumps(body))
+            return DBResponse(response_code=500, body=body)
     elif isinstance(exception, psycopg2.errors.UniqueViolation):
-        return DBResponse(response_code=409, body=json.dumps(body))
+        return DBResponse(response_code=409, body=body)
     elif isinstance(exception, IndexError):
         return DBResponse(response_code=404, body={})
     else:
-        return DBResponse(response_code=500, body=json.dumps(body))
```
oavdeev, line 27: Unless I'm missing something, before this fix body was doubly-encoded here https://github.com/Netflix/metaflow-service/blob/master/services/metadata_service/api/utils.py#L24

romain-intel: This is somewhat duplicative of #184 (or the idea is for that to be the cleanup PR). I need to change things in the client too and we need to support both ways of doing things. But yes, this is planned.

PR #201, opened 2021-07-13
fix: solving cache action boto3 issues

saikonen:
rewrite S3 fetches to use vanilla boto3 instead of the metaflow client implementation. Add logic to correctly check filesize before starting S3 downloads, in order to skip files that are too big.

cleaned up unnecessary S3 client from the codebase


msavela (approved):
This looks really good!

PR #203, opened 2021-07-14
feat[s3]: Customize S3 client endpoint url as well as SSL verify flag

msavela:
Two new environment variables for UI service:

- `METAFLOW_S3_ENDPOINT_URL`
- `METAFLOW_S3_VERIFY_CERTIFICATE`

This change is similar to Metaflow boto3 client init https://github.com/Netflix/metaflow/blob/dc42d670e6c3f7555f40290fce6ed46eb0c12b17/metaflow/datastore/util/s3util.py#L19


saikonen (approved):

saikonen:
Approved for the cache action changes, excellent stuff. Do note that the API handlers use aiobotocore for the log routes, so this needs its own `get_s3_client()` if you intend on testing that functionality as well.

saikonen (approved):

PR #211, opened 2021-09-07
Pinning Go Dep in docker file.

valayDave:
Pinning the go docker dependency otherwise the image is not building; Ran into some issue with `go:latest` image so pinning the working image.

Issue #152, opened 2021-04-13
Metadata DB Password should be quoted


displacedaussie:
When launching a Metaflow stack using the recommended [CloudFormation template](https://github.com/Netflix/metaflow-tools/blob/master/aws/cloudformation/metaflow-cfn-template.yml#L541), an RDS database password is created using AWS Secrets Manager. This is currently configured to exclude the characters `"@/\`.

Unfortunately, this still allows potentially problematic characters like `#` and `'` to be included in the password.

If `'` is included as the _first_ character in the password, the Metadata Service fails to start as follows:

```
/migration_service/migration_server.py:17: DeprecationWarning: loop argument is deprecated
  app = web.Application(loop=loop)
printing connection exception: invalid dsn: unterminated quoted string in connection info string

printing connection exception: invalid dsn: unterminated quoted string in connection info string

printing connection exception: invalid dsn: unterminated quoted string in connection info string

/migration_service/migration_server.py:28: DeprecationWarning: Application.make_handler(...) is deprecated, use AppRunner API instead
  handler = the_app.make_handler()
serving on ('0.0.0.0', 8082)
Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/aiohttp/web_protocol.py", line 418, in start
    resp = await task
  File "/usr/local/lib/python3.7/dist-packages/aiohttp/web_app.py", line 458, in _handle
    resp = await handler(request)
  File "/migration_service/api/admin.py", line 53, in version
    version = await ApiUtils.get_latest_compatible_version()
  File "/migration_service/api/utils.py", line 56, in get_latest_compatible_version
    is_present = await PostgresUtils.is_present("flows_v3")
  File "/migration_service/data/postgres_async_db.py", line 10, in is_present
    with (await AsyncPostgresDB.get_instance().pool.cursor()) as cur:
AttributeError: 'NoneType' object has no attribute 'cursor'
/bin/sh: 1: metadata_service: not found
```

The password should be escaped when [setting up the DSN](https://github.com/Netflix/metaflow-service/blob/master/services/data/postgres_async_db.py#L59), otherwise the service can fail to start, resulting in a Metaflow stack that does not work.

oavdeev:
[...]

PR #212, opened 2021-09-07
Quoted password fix

valayDave:
This fixes the issue noted #152 . To merge this we first need to merge  #211  as it ensures that the docker image builds without any problems.

TODO: Need to find a better way to escape special chars over the REGEX-based approach to escaping special characters.

### DIFF
--- a/services/migration_service/migration_config.py
+++ b/services/migration_service/migration_config.py
@@ -1,9 +1,17 @@
 from services.utils import DBConfiguration
-
+# Shell Related Password and username setting happens here for Goose. 
+# Ensure we escape the right characters in username/password to avoid goose failures. 
+SHELL_ESCAPE_TRANSLATION = str.maketrans({"-":  r"\-",
+                "]":  r"\]",
+                "\\": r"\\",
+                "^":  r"\^",
+                "$":  r"\$",
+                "*":  r"\*",
+                ".":  r"\."})
 db_conf = DBConfiguration()
 
 host = db_conf.host
 port = db_conf.port
-user = db_conf.user
-password = db_conf.password
+user = db_conf.user.translate(SHELL_ESCAPE_TRANSLATION)
+password = db_conf.password.translate(SHELL_ESCAPE_TRANSLATION)
 database_name = db_conf.database_name
--- a/services/utils/__init__.py
+++ b/services/utils/__init__.py
@@ -6,6 +6,7 @@ from urllib.parse import urlencode
 from aiohttp import web
 from typing import Dict
 import logging
+import psycopg2
 
 # The latest commit hash of the repository, if set as an environment variable.
 
@@ -74,6 +75,7 @@ def format_baseurl(request: web.BaseRequest):
 #   3. Env connection arguments (MF_METADATA_DB_HOST="..." MF_METADATA_DB...)
 #   4. Default connection arguments (DBConfiguration(host="..."))
 #
+ 
 class DBConfiguration(object):
     host: str = None
     port: int = None
@@ -89,7 +91,6 @@ class DBConfiguration(object):
     timeout: int = None  # aiopg default: 60 (seconds)
 
     _dsn: str = None
-
     def __init__(self,
                  dsn: str = None,
                  host: str = "localhost",
@@ -101,24 +102,63 @@ class DBConfiguration(object):
                  pool_min: int = 1,
                  pool_max: int = 10,
                  timeout: int = 60):
-        table = str.maketrans({"'": "\'", "`": r"\`"})
 
         self._dsn = os.environ.get(prefix + "DSN", dsn)
-
-        self.host = os.environ.get(prefix + "HOST", host).translate(table)
-        self.port = int(os.environ.get(prefix + "PORT", port))
-        self.user = os.environ.get(prefix + "USER", user).translate(table)
-        self.password = os.environ.get(
-            prefix + "PSWD", password).translate(table)
-        self.database_name = os.environ.get(
-            prefix + "NAME", database_name).translate(table)
+        # We explicitly check the validity of the DSN string 
+        # to avoid issues caused by bad DSN strings. 
+        if not self._is_valid_dsn(self._dsn):
+            self._dsn = None
+        self._host = os.environ.get(prefix + "HOST", host)
+        self._port = int(os.environ.get(prefix + "PORT", port))
+        self._user = os.environ.get(prefix + "USER", user)
+        self._password = os.environ.get(prefix + "PSWD", password)
+        self._database_name = os.environ.get(prefix + "NAME", database_name)
 
         self.pool_min = int(os.environ.get(prefix + "POOL_MIN", pool_min))
         self.pool_max = int(os.environ.get(prefix + "POOL_MAX", pool_max))
 
         self.timeout = int(os.environ.get(prefix + "TIMEOUT", timeout))
+    
+    @staticmethod
+    def _is_valid_dsn(dsn):
+        if dsn is None:
+            return False
+        try:
+            psycopg2.extensions.parse_dsn(dsn)
+            return True
+        except psycopg2.ProgrammingError: 
+            # This means that the DSN is unparsable. 
+            return False
 
     @property
     def dsn(self):
-        return self._dsn or "dbname={0} user={1} password={2} host={3} port={4}".format(
-            self.database_name, self.user, self.password, self.host, self.port)
+        if self._dsn is None:
+            return psycopg2.extensions.make_dsn(
+                dbname=self._database_name ,
+                user=self._user,
+                host=self._host, 
+                port=self._port,
+                password=self._password
+            )
+        else:
+            return self._dsn
+
+    @property
+    def port(self):
+        return self._port
+
+    @property
+    def password(self):
+        return self._password
+    
+    @property
+    def user(self):
+        return self._user
+    
+    @property
+    def database_name(self):
+        return self._database_name
+    
+    @property
+    def host(self):
+        return self._host
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:1.16.3
 
 RUN go get -u github.com/pressly/goose/cmd/goose
 
--- a/services/utils/tests/unit_tests/utils_test.py
+++ b/services/utils/tests/unit_tests/utils_test.py
@@ -56,7 +56,7 @@ def test_db_conf():
 
 def test_db_conf_dsn():
     with set_env():
-        assert DBConfiguration(dsn='foo').dsn == 'foo'
+        assert DBConfiguration(dsn='user=foo').dsn == 'user=foo'
 
 
 def test_db_conf_arguments():


oavdeev:
I suggest doing something like `shlex.join(['goose', 'postgres', make_dsn(...), ...] ` to generate `goose_template`

services/migration_service/migration_config.py
```diff
@@ -1,9 +1,17 @@
 from services.utils import DBConfiguration
-
```
valayDave, line 2: @oavdeev We can now remove this all together; Using URL encoded strings makes this work.

services/utils/__init__.py
```diff
@@ -101,24 +102,68 @@ def __init__(self,
...
+            return False
+        try:
+            psycopg2.extensions.parse_dsn(dsn)
+            return True
+        except psycopg2.ProgrammingError: 
+            # This means that the DSN is unparsable. 
+            return False
+    
+    @property
+    def connection_string(self):
```
valayDave, line 134: > I suggest doing something like `shlex.join(['goose', 'postgres', make_dsn(...), ...] ` to generate `goose_template`

Circumvented `shlex` via URL encoding the PG connection string; This makes passing it to shell very simple for goose.

services/migration_service/migration_config.py
```diff
@@ -1,9 +1,2 @@
...
-db_conf = DBConfiguration()
-
```
valayDave, line 4: @oavdeev I have removed the exposure of host, port, user, and password by exposing a single db_conf object that has a URL encoded connection string that can be given to goose.

services/utils/__init__.py
```diff
@@ -101,24 +102,68 @@ def __init__(self,
                  pool_min: int = 1,
                  pool_max: int = 10,
                  timeout: int = 60):
-        table = str.maketrans({"'": "\'", "`": r"\`"})
```
valayDave, line 104: Removing this for a more general-purpose translation of user, password, host, and port via Postgres connection string and DSN; The DSN doesn't require escaping when we give to `aiopg` ; For [goose](https://github.com/pressly/goose#usage) we provide PG connection string which is supported after [Postgres version 9.2](https://www.postgresql.org/docs/9.2/libpq-connect.html#LIBPQ-CONNSTRING).

services/utils/__init__.py
oavdeev, line 134: naming nit: i think often "connection string" and DSN are used interchangeably but here it looks like they refer to different thing

services/utils/__init__.py
```diff
@@ -101,24 +102,68 @@ def __init__(self,
...
-        self.host = os.environ.get(prefix + "HOST", host).translate(table)
-        self.port = int(os.environ.get(prefix + "PORT", port))
-        self.user = os.environ.get(prefix + "USER", user).translate(table)
-        self.password = os.environ.get(
-            prefix + "PSWD", password).translate(table)
-        self.database_name = os.environ.get(
-            prefix + "NAME", database_name).translate(table)
+        # We explicitly check the validity of the DSN string 
+        # to avoid issues caused by bad DSN strings. 
+        if not self._is_valid_dsn(self._dsn):
```
oavdeev, line 109: if the DSN is set but invalid, wouldn't it be better to just fail loudly?

services/migration_service/api/admin.py
```diff
@@ -69,7 +69,7 @@ async def upgrade(self, request):
                 description: could not upgrade
         """
         goose_version_cmd = goose_migration_template.format(
-            database_name, user, password, host, port,
+            db_conf.connection_string,
             "up"
```
oavdeev, line 73: It would still be a good idea to do `shlex.join` (or Py3.7 equivalent) here instead of `.format`

services/utils/__init__.py
```diff
@@ -101,24 +102,68 @@ def __init__(self,
...
+    def user(self):
+        return self._user
+    
+    @property
+    def database_name(self):
+        return self._database_name
+    
+    @property
+    def host(self):
+        return self._host
```
oavdeev, line 169: are these properties used anywhere?

oavdeev (changes requested):

services/utils/__init__.py
valayDave: In the unit tests : https://github.com/Netflix/metaflow-service/blob/master/services/utils/tests/unit_tests/utils_test.py#L64

oavdeev: Ah I see. Now I'm not sure what does this test tests though, maybe we should remove it

services/migration_service/api/admin.py
valayDave: Done.

services/utils/__init__.py
valayDave: Done.

services/migration_service/api/admin.py
valayDave: Done

services/utils/__init__.py
valayDave: Done.

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
-    "port={} " \
-    "sslmode=disable\" {}"
+
+def make_goose_migration_template(conn_str,command):
+    return ' '.join(shlex.quote(arg) for arg in [
+        "goose",\
+        "-dir",\
+        path,\
+        "postgres",\
+        f"\"{conn_str}\"",\
```
oavdeev, line 29: i don't think you need the quotes here, shlex will add them if necessary
```suggestion
        conn_str,
```

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
+        f"{command}"
```
oavdeev, line 30:
```suggestion
        command
```

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
-                 "user={} " \
-                 "password={} " \
-                 "host={} " \
-                 "port={} " \
-                 "sslmode=disable\" {}"
+def make_goose_template(conn_str,command):
+    return ' '.join(shlex.quote(arg) for arg in [
+        "goose",\
+        "postgres",\
+        f"\"{conn_str}\"",\
```
oavdeev, line 17:
```suggestion
        conn_str,
```

services/utils/__init__.py
valayDave: Can we scope this change in a new PR?

oavdeev (approved):

services/migration_service/api/__init__.py
valayDave: Lemme test that

valayDave: Done.

services/migration_service/api/__init__.py
valayDave: Test works.

services/utils/__init__.py
oavdeev: oh yeah sorry i didn't mean in this one

services/utils/__init__.py
```diff
@@ -101,24 +108,68 @@ def __init__(self,
...
 
         self.timeout = int(os.environ.get(prefix + "TIMEOUT", timeout))
+    
+    @staticmethod
+    def _is_valid_dsn(dsn):
+        if dsn is None:
+            return False
+        try:
+            psycopg2.extensions.parse_dsn(dsn)
+            return True
```
oavdeev, line 133: nit: it looks like the return value isn't used anywhere

services/utils/__init__.py
```diff
@@ -17,6 +18,12 @@
 log_level = os.environ.get('LOGLEVEL', 'INFO').upper()
 logging.basicConfig(level=log_level)
 
+class InvalidDSNException(Exception):
+    headline = 'Invalid DSN String'
+    def __init__(self, lineno=None):
+        self.message = f"The DSN String Provided is Invalid"
+        self.line_no = lineno
```
oavdeev, line 25: what is `lineno` for?

oavdeev: nit: i think it would be perfectly fine to just `raise Exception("Invalid DSN string")` and not have a separate exception class, as we don't catch this specific exception anywhere anyway

valayDave: Will Fix.

services/utils/__init__.py
valayDave: Will Fix.

services/utils/__init__.py
valayDave: `lineno` was by mistake; I will refactor your solution and update the PR.

valayDave: Done.

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
-goose_template = "goose postgres " \
-                 "\"dbname={} " \
-                 "user={} " \
-                 "password={} " \
-                 "host={} " \
-                 "port={} " \
-                 "sslmode=disable\" {}"
+def make_goose_template(conn_str,command):
+    return ' '.join(shlex.quote(arg) for arg in [
+        "goose",\
```
romain-intel, line 15: Nit: backslashes are redundant.

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
 latest = "latest"
 
-goose_template = "goose postgres " \
-                 "\"dbname={} " \
-                 "user={} " \
-                 "password={} " \
-                 "host={} " \
-                 "port={} " \
-                 "sslmode=disable\" {}"
+def make_goose_template(conn_str,command):
```
romain-intel, line 13: Nit: space after comma.

services/migration_service/api/__init__.py
```diff
@@ -9,19 +10,22 @@
...
 path = os.path.dirname(__file__) + "/../migration_files"
-goose_migration_template = "goose -dir " + path + " postgres " \
-    "\"dbname={} " \
-    "user={} " \
-    "password={} " \
-    "host={} " \
-    "port={} " \
-    "sslmode=disable\" {}"
+
+def make_goose_migration_template(conn_str,command):
```
romain-intel, line 23: Ditto for here (comma + backslash)

services/migration_service/api/utils.py
```diff
@@ -25,10 +26,7 @@ def get_unapplied_migrations(current_version):
     @staticmethod
     async def get_goose_version():
         # if tables exist but goose doesn't find version table then
-        goose_version_cmd = goose_template.format(
-            database_name, user, password, host,
-            port, "version"
-        )
+        goose_version_cmd = make_goose_template(db_conf.connection_string_url,'version')
```
romain-intel, line 29: space after comma.

services/migration_service/api/utils.py
```diff
@@ -57,20 +55,16 @@ async def get_latest_compatible_version():
             version = await ApiUtils.get_goose_version()
             return version_dict[version]
         else:
-            goose_version_cmd = goose_migration_template.format(
-                database_name, user, password, host, port,
-                "up"
-            )
+            goose_version_cmd = make_goose_migration_template(db_conf.connection_string_url,'up')
```
romain-intel, line 58: Ditto

services/migration_service/api/utils.py
```diff
@@ -57,20 +55,16 @@ async def get_latest_compatible_version():
...
             p.wait()
             return latest
 
     @staticmethod
     async def is_migration_in_progress():
-        goose_version_cmd = goose_template.format(
-            database_name, user, password, host,
-            port, "status"
+        goose_version_cmd = make_goose_template(
+            db_conf.connection_string_url,"status"
```
romain-intel, line 67: ditto

services/utils/__init__.py
```diff
@@ -2,10 +2,11 @@
 import sys
 import os
 import traceback
-from urllib.parse import urlencode
+from urllib.parse import urlencode,quote
```
romain-intel, line 5: ditto

services/utils/__init__.py
```diff
@@ -101,24 +101,67 @@ def __init__(self,
...
-        self.host = os.environ.get(prefix + "HOST", host).translate(table)
-        self.port = int(os.environ.get(prefix + "PORT", port))
-        self.user = os.environ.get(prefix + "USER", user).translate(table)
-        self.password = os.environ.get(
-            prefix + "PSWD", password).translate(table)
-        self.database_name = os.environ.get(
-            prefix + "NAME", database_name).translate(table)
+        # We explicitly check the validity of the DSN string 
+        # to avoid issues caused by bad DSN strings. 
+        self._is_valid_dsn(self._dsn)
```
romain-intel, line 108: Confused here: the test will fail with a `None` `dsn`. Do we now require one? I'd be OK if it ignored the None value.

services/utils/__init__.py
```diff
@@ -101,24 +101,67 @@ def __init__(self,
...
+            raise Exception("Invalid DSN String")
+        
+    
+    @property
+    def connection_string_url(self):
+        # postgresql://[user[:password]@][host][:port][/dbname][?param1=value1&...]
+        return f'postgresql://{quote(self._user)}:{quote(self._password)}@{self._host}:{self._port}/{self._database_name}?sslmode=disable'
 
     @property
     def dsn(self):
```
romain-intel, line 137: Maybe in init, if dsn is null, we could compute it and validate it and then use that?

romain-intel (commented):
A few comments (minor) and a question about dsn (maybe bigger)

PR #215, opened 2021-09-14
fix: streamline run statuses and remove fallbacks

saikonen:
- remove unnecessary fallback to end step _task_ok artifact for run status inference
- rely solely on run heartbeat and attempt/attempt_ok metadata entries for status, duration and finished_at timestamp
- update integration tests to reflect changes
- Also solves incorrect status for runs during end-step retries.

### DIFF
--- a/services/ui_backend_service/data/db/tables/run.py
+++ b/services/ui_backend_service/data/db/tables/run.py
@@ -66,22 +66,6 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
             table_name=table_name,
             metadata_table=metadata_table
         ),
-        """
-        LEFT JOIN LATERAL (
-            SELECT ts_epoch
-            FROM {task_table} as task
-            WHERE
-                task.flow_id={table_name}.flow_id
-                AND task.run_number={table_name}.run_number
-                AND task.last_heartbeat_ts IS NULL
-                AND end_attempt IS NULL
-            ORDER BY ts_epoch DESC
-            LIMIT 1
-        ) as pending_task ON true
-        """.format(
-            table_name=table_name,
-            task_table=task_table
-        ),
     ]
 
     @property
@@ -106,7 +90,6 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
             WHEN end_attempt_ok.ts_epoch IS NOT NULL
             THEN end_attempt_ok.ts_epoch
             WHEN {table_name}.last_heartbeat_ts IS NOT NULL
-                AND pending_task IS NOT NULL
                 AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)>{heartbeat_threshold}
             THEN {table_name}.last_heartbeat_ts*1000
             ELSE NULL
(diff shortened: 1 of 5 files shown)


services/ui_backend_service/data/db/tables/run.py
```diff
@@ -22,34 +23,83 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
...
+                    task.step_name=attempt_ok.step_name AND
+                    task.task_id=attempt_ok.task_id AND
+                    attempt_ok.field_name='attempt_ok'
+                LIMIT 1
+            ) AS attempt_ok ON true
+            WHERE
+                {table_name}.flow_id = task.flow_id
+                AND {table_name}.run_number = task.run_number
+                AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)>{heartbeat_threshold}
+                AND end_attempt_ok IS NULL
```
romain-intel, line 89: I *think* you may be able to filter on `end_attempt IS NULL OR end_attempt_ok IS NULL` which should eliminate more checks (ie: as soon as `end` starts, it won't check this as opposed to `wait until we see an `attempt_ok` for `end`. Not sure it makes a huge difference but maybe?

services/ui_backend_service/data/db/tables/run.py
```diff
@@ -22,34 +23,83 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
...
+                AND (
+                    (attempt_ok.is_ok IS FALSE) OR /* failed attempt_ok */
+                    (attempt_ok.is_ok IS NOT TRUE AND @(extract(epoch from now())-task.last_heartbeat_ts)>{heartbeat_threshold}) /* failed heartbeat */
```
romain-intel, line 92: Shouldn't this be something like `attempt_ok IS NULL`? Wondering about the case when there is no `attempt_ok` metadata and the heartbeat is stale. Maybe the `attempt_ok.is_ok IS NOT TRUE` covers that though. Not 100% sure.

services/ui_backend_service/data/db/tables/run.py
```diff
@@ -22,34 +23,83 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
...
+                )
+            GROUP BY task.flow_id, task.run_number, task.step_name, task.task_id, attempt_ok.ts_epoch
+            ORDER BY attempt_ok.ts_epoch DESC
```
romain-intel, line 95: OK, so this will select effectively the task that missed a heartbeat (any of those) followed by the last task to have failed (if none missed a heartbeat). I think that works.

romain-intel (commented):
Just a few questions but I like the logic :).

services/ui_backend_service/data/db/tables/run.py
saikonen: yeah that was the intent so the second case would catch both is_ok being false and null. Only checking for null should also work actually, I'll try it out.

services/ui_backend_service/data/db/tables/run.py
saikonen: added the end_attempt IS NULL to the filter as well, didn't seem to have a significant impact though, but makes sense to take it along for the ride.

romain-intel: I would have added as an OR but ya, it probably isn't hugely significant.

msavela (commented):
LGTM

msavela (approved):

PR #221, opened 2021-09-27
Update run heartbeat on task heartbeat

romain-intel:


services/data/postgres_async_db.py
```diff
@@ -299,10 +301,18 @@ async def update_row(self, filter_dict={}, update_dict={}):
...
+            operator = '='
             v = str(col_val).strip("'")
             if not v.isnumeric():
                 v = "'" + v + "'"
-            filters.append(col_name + "=" + str(v))
+            find_operator = operator_match.match(col_name)
+            if find_operator:
+                col_name = find_operator.group(1)
+                operator = find_operator.group(2)
+                filters.append('(%s IS NULL or %s %s %s)' %
```
msavela, line 312: It seems that this existed before current implementation, but it is recommended to pass variables separately via `execute()` instead of string interpolation. This might not be a problem since the service is usually running inside trusted network, but better safe than sorry.

msavela: More details here: https://www.psycopg.org/docs/usage.html#passing-parameters-to-sql-queries

msavela (approved):

services/data/postgres_async_db.py
romain-intel: Yes, this was the case before. I can make a separate PR to harden this. It's a good point and should be done. I just followed the style of how things were for this change.

PR #223, opened 2021-10-01
Cache: Metaflow Client API support

msavela:
This PR introduces support for Metaflow Client API inside Cache Actions.

**Currently implemented**
- Task statuses
- Run parameters
- DAG (codepackage) fetching
- Artifact content fetching
- Artifact search
- Preloading runs (Parameters + Task statuses + DAG)
- Realtime data postprocessing

**What’s still missing**
- Attempt id support for Tasks & Artifacts
  - Already supported by Metaflow+Service, needs to be implemented here
- Artifact size check (HEAD request) before fetching the actual data
  - No support on Metaflow client
- Tests broken

UI Service now requires environment variable `METAFLOW_SERVICE_URL` to be defined in order to communicate with Metaflow Client API.

Depends on https://github.com/Netflix/metaflow-service/pull/216


services/ui_backend_service/data/cache/get_task_action.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class GetTask(GetData):
+    @classmethod
+    def format_request(cls, pathspecs: List[str], invalidate_cache=False):
+        return super().format_request(targets=pathspecs, invalidate_cache=invalidate_cache)
+
+    @classmethod
+    def fetch_data(cls, pathspec):
+        task = Task(pathspec)
```
saikonen, line 15: still todo: implement error wrapping for the data fetching, and streaming it out to the action as previously.

the rewrite looks excellent though, very clean! :)

services/ui_backend_service/data/cache/get_parameters_action.py
```diff
@@ -0,0 +1,26 @@
+from typing import List
+
+from .get_data_action import GetData
+
+from metaflow import Step
+
+
+class GetParameters(GetData):
+    @classmethod
+    def format_request(cls, pathspecs: List[str], invalidate_cache=False):
```
saikonen, line 10: maybe naming could be specific, as this will only work for **run** pathspecs due to the appending of `_parameters` to get a step?

services/ui_backend_service/data/cache/generate_dag_action.py
```diff
@@ -134,8 +112,3 @@ def generate_dag(flow_id, tarball_path):
             'doc': node.doc
         }
     return dag
-
-
-class UnsupportedFlowLanguage(Exception):
-    def __str__(self):
-        return "Parsing DAG graph is not supported for the language used in this Flow."
```
saikonen, line 141: did this issue get alleviated by using the metaflow client? were you able to generate a DAG for a flow written with R?

services/ui_backend_service/data/cache/client/cache_worker.py
```diff
@@ -5,6 +5,10 @@
 
 from .cache_action import import_action_class_spec
 
+os.environ['METAFLOW_S3_RETRY_COUNT'] = '0'
+os.environ['USERNAME'] = 'none'
+os.environ['METAFLOW_DEFAULT_METADATA'] = 'service'
+
```
saikonen, line 11: possibly best to keep these configurable, and default to the values here?

services/ui_backend_service/data/cache/get_task_action.py
saikonen: I'm implementing a decorator for the log feature which will wrap the S3 errors raised by the metaflow library, and use the existing custom errors with corresponding id's so this is already on its way.

what seems to be needed in this case is access to the `stream_output` from `execute()` inside fetch_data to be able to stream errors.

services/ui_backend_service/api/notify.py
```diff
@@ -100,7 +100,7 @@ async def handle_trigger_msg(self, msg: str):
                 if operation == "INSERT" and \
                         table.table_name == self.db.step_table_postgres.table_name and \
                         data["step_name"] == "start":
-                    self.event_emitter.emit("run-parameters", data['flow_id'], data['run_number'])
+                    self.event_emitter.emit("preload-run-parameters", data['flow_id'], data['run_number'])
```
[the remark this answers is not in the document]
msavela: Preloading Run parameters is now part of the process here: https://github.com/Netflix/metaflow-service/pull/223/files#diff-17bee700a8b2eb91cc38dcb3d7e30cd88769966fdb21c724393a5746f664d3cfR153

And in both cases I would consider this to be "preloading", where parameters are loaded before

services/ui_backend_service/data/cache/get_task_action.py
msavela: Added `stream_output` to fetch_data.

services/ui_backend_service/data/cache/get_parameters_action.py
msavela: Added more documentation on input parameters

PR #233, opened 2021-10-18
feat: add ack responses to websocket subscriptions

saikonen:
Adds `ACK` responses to web socket resource subscriptions. Used for timing the sending of GET requests only after a WS subscription is active on the UI, so no messages get lost in-between.

The feature is meant to deal with a rare edge-case when running flows with tasks that complete extremely fast. In rare cases the user can navigate to the timeline view, and between the `GET` request and web socket subscription happening, some `completed` status broadcasts can be missed. 
Proposed change is to be deliberate in the order of operations, where the `GET` follows only a successful subscription to a resource.

***Breaking Change***
current UI does not handle `ACK` types at all and will break with this, only merge at the same time the corresponding UI changes go in.

### DIFF
--- a/services/ui_backend_service/api/ws.py
+++ b/services/ui_backend_service/api/ws.py
@@ -144,6 +144,12 @@ class Websocket(object):
                 self.loop.create_task(
                     await self._event_subscription(subscription, event['operation'], event['resources'], event['data'])
                 )
+        # subscribed successfully, send ACK
+        payload = {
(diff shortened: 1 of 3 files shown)


saikonen:
Moving back to Draft for the time being, as work is still ongoing on the UI counterpart for this PR.

saikonen:
closing this for now as it is a major breaking change for the UI support, and frontend work has not been progressing on the issue for now.

PR #254, opened 2021-10-29
Feat/expose timestamps for loglines

saikonen:
Returns timestamps for log lines as part of the API response when they are available. Frontend can choose to display, or ignore them.

Slight concern that different logstyle access' implementation details are leaking from the metaflow cli into the cache action, but this has to be done in order to access (timestamp, logline) tuple stream from MFLOGs, and at the same time support old style logs as well.


romain-intel:
we can merge this whenever you want.

PR #257, opened 2021-11-02
fix: pin test Dockerfile golang dependency

saikonen:
golang:latest has issues with the goose installation for whatever reason. Pinning the version fixes this and the docker image build successfully.

Change is more in line with the other Dockerfiles also having the version pinned.


romain-intel:
Looks like it was the same as: https://github.com/Netflix/metaflow-service/commit/a7bc059f641d759bd7160c54958e2d5b74cd562e

PR #258, opened 2021-11-03
chore: unpin metaflow version from ui_backend_service

saikonen:
Unpin metaflow client version from ui_service requirements, as the service should always work with latest client version.

Keep a lower bound to guard against installing unsupported versions though.


romain-intel (approved):
LGTM

PR #261, opened 2021-11-05
feat: more generic cache error handling

saikonen:
generalise the error passing from cache actions to api responses. Favour passing the actual exception classname as an error id, to accommodate possible additions to exceptions that the client can raise.

implements a context manager for streaming errors in a standardised way from cache actions to stream_output.

cleanup unused code from cache implementation


romain-intel (approved):
I'm going to trust this works (seems to on our instance :))

PR #268, opened 2021-11-24
feat: support ui_backend configuration with json

saikonen:
Add another method of configuring optional components of the service, as passing massive json configuration as environment variables can become difficult to handle.

The following environment variable configs now have an alternative file config available
- `NOTIFICATIONS` - `config.notifications.json`
- `CUSTOM_QUICKLINKS` - `config.custom_quicklinks.json`
- `PLUGINS` - `config.plugins.json`

This change should not have any adverse effect on existing deployments configured with environment variables. The json file lookup is treated as a fallback only in case an env var is not set.


services/ui_backend_service/api/utils.py
```diff
@@ -20,6 +45,14 @@ def get_json_from_env(variable_name: str):
         return None
 
 
+def get_json_from_file(filepath: str):
+    try:
+        with open(filepath) as f:
+            return json.load(f)
+    except Exception:
+        return None
```
obgibson, line 53: In #283 I added more logging into when there is an exception. I think that would be helpful here too

services/ui_backend_service/docs/plugins.md
```diff
@@ -5,7 +5,7 @@ There are two ways to register a new plugin to UI service:
 1. Placing a folder under `services/ui_backend_service/plugins/installed`
 2. Defining a remote Git repository location
 
-Plugins can be registered via `PLUGINS` environment variable. The value should be a _stringified_ json of the format:
+Plugins can be registered via the `PLUGINS` environment variable or with a `config.plugins.json` file on the server. See `example.plugins.json` for reference. The value of the environment variable should be a _stringified_ json of the format:
```
obgibson, line 8: `stringified` is a bit confusing and inaccurate for the JSON file. I would spell out that the env var needs to escape `"` and spaces with a `\` and the config file can just be regular old JSON.

obgibson (changes requested):
Tested for plugins, both as env and config file and it worked.

services/ui_backend_service/api/utils.py
saikonen: Added some comments to the other PR regarding logging. Added logging to the config file loader here as suggested, with the caveat that we're only interested in cases where a file is loaded (found), but something goes wrong.

a config file not existing, or a config env_var not being set is not a reportable incident in my opinion, as these are optional configs

obgibson (approved):

PR #271, opened 2021-12-16
feat: Support client generated DAG

saikonen:
Add support for the metaflow client generated DAG https://github.com/Netflix/metaflow/pull/822
- Defaults to trying to fetch the `_graph_info` artifact if it exists and directly outputting it for DAG rendering.
- Fallback to previous `generate_dag` implementation. Reworked this to conform to the new graph_info format as well.

Note: Depends on https://github.com/Netflix/metaflow-ui/pull/45


services/ui_backend_service/data/cache/custom_flowgraph.py
```diff
@@ -210,25 +202,47 @@ def __str__(self):
...
-                      '  shape = "record" ];'.format(node, type=nodetype)
-
-        return "digraph {0.name} {{\n"\
-               "rankdir=LR;\n"\
-               "{nodes}\n"\
-               "{edges}\n"\
-               "}}".format(self,
-                           nodes='\n'.join(node_specs()),
-                           edges='\n'.join(edge_specs()))
+    def output_steps(self):
```
saikonen, line 205: Would be preferable to completely get rid of the custom flowgraph implementation from the backend, and use the client library's FlowGraph  class for parsing a DAG even when only a codepackage is present.

The client flowgraph does not accept a `run.code.flowspec` directly though, and maybe it would be better for the client to simply have a property for `_graph_info` that has clientside fallback logic which uses the codepackage, in case the artifact is missing.

services/ui_backend_service/data/cache/generate_dag_action.py
```diff
@@ -86,8 +87,17 @@ def execute(cls,
         result_key = [key for key in keys if key.startswith('dag:result')][0]
 
         with streamed_errors(stream_output):
+            # TODO: Clean up the internal _graph_info accessing.
+            # Having to jump through a lot of hoops currently due to internal artifacts
```
saikonen, line 91: easiest change is if a new client version provides a property on the run, along with fallback logic that parses graph_info from codepackage if the artifact is missing.

obgibson (approved):
Functionality tested with https://github.com/Netflix/metaflow-ui/pull/45

services/ui_backend_service/api/dag.py
```diff
@@ -85,6 +72,68 @@ async def get_run_dag(self, request):
...
 
+    async def get_run_ids_from_codepackage_metadata(self, request):
+        """
+        Tries to locate 'code-package' or 'code-package-url' in run metadata.
+
+        Returns
+        -------
+        Tuple
+            flow_id : str
+            run_id : str or int
```
oavdeev, line 83: nit: it doesn't really return a tuple, right?

services/ui_backend_service/api/dag.py
```diff
@@ -43,24 +43,11 @@ async def get_run_dag(self, request):
...
-                flow_name, run_id_value,
-                "code-package", "code-package-url"
-            ],
-            fetch_single=True, expanded=True
-        )
+        db_response = await self.get_run_ids_from_graph_info_artifact(request)
+        if not db_response.response_code == 200:
+            # Try to look for codepackage if graph artifact is missing
+            db_response = await self.get_run_ids_from_codepackage_metadata(request)
+
```
oavdeev, line 50: Hmm I don't entirely follow the logic here, can you add a comment explaining it? 

I _think_ the original version of this code would essentially simplify verify that the run was executed on a remote backend and has a code package (otherwise getting constructing the graph would be impossible). Now we want to check that either the run has `_graph_info` or has the code package.

If that's true, worth adding a comment explaining all that and maybe naming the functions something like `check_run_has_graph_info` or similar, since its not literally getting run_id from the artifact here

services/ui_backend_service/data/cache/custom_flowgraph.py
```diff
@@ -210,25 +202,58 @@ def __str__(self):
...
+
+        steps_info = {}
+        graph_structure = []
+
+        def node_to_type(node):
```
oavdeev, line 210: I'd add a comment explaining why the return value is different from `node.type` and what it used for later

services/ui_backend_service/api/dag.py
```diff
@@ -85,6 +72,68 @@ async def get_run_dag(self, request):
...
+    async def get_run_ids_from_graph_info_artifact(self, request):
+        """
+        Tries to locate '_graph_info' in run artifacts
+
+        Returns
+        -------
+        Tuple
+            flow_id : str
+            run_id : str or int
+        """
```
oavdeev, line 115: doesn't seem to return a tuple here either

oavdeev (changes requested):

oavdeev:
Overall lgtm! just a couple of nitpicks for readability

Also it seems like we don't have tests for this stuff? How hard would it be to add them (could be a separate PR)

services/ui_backend_service/api/dag.py
saikonen: cleaned up dag api router a bit, including the type notation.

services/ui_backend_service/data/cache/custom_flowgraph.py
saikonen: > Overall lgtm! just a couple of nitpicks for readability
> 
> Also it seems like we don't have tests for this stuff? How hard would it be to add them (could be a separate PR)

Unfortunately at this moment the cache is not a part of any integration end-to-end testing, there are only unit tests on the cache action level, and in case a cache response is required for other testing, we have usually used patching

services/ui_backend_service/api/dag.py
saikonen: cleaned up now.

services/ui_backend_service/api/dag.py
saikonen: cleaned up the structure a bit and added comments. You're correct in the general logic, we want to check before firing the cache action, whether the required data for a DAG actually exists. Now there is a clear fallback where we check for `_graph_info` artifact -> `code-package` metadata -> 404 if neither exist.

services/ui_backend_service/data/cache/custom_flowgraph.py
saikonen: added a comment to the function.

oavdeev (approved):

PR #272, opened 2021-12-17
Gracefully handling db failure for `/artifacts` API

valayDave:
## Context 

Users of metaflow after upgrading the service were getting `500` Errors when they try to access Data for task related to some flow; They managed to access metadata though. Slack Thread with more context : https://outerboundsco.slack.com/archives/C02116BBNTU/p1636723735291800

The Debug logs from the MD service looked like the following :
```
2021-12-10 08:13:18.782,ERROR:aiohttp.server:Error handling request
2021-12-10 08:13:18.782,Traceback (most recent call last):
2021-12-10 08:13:18.782,"  File ""/opt/latest/lib/python3.7/site-packages/aiohttp/web_protocol.py"", line 435, in _handle_request"
2021-12-10 08:13:18.782,    resp = await request_handler(request)
2021-12-10 08:13:18.782,"  File ""/opt/latest/lib/python3.7/site-packages/aiohttp/web_app.py"", line 504, in _handle"
2021-12-10 08:13:18.782,    resp = await handler(request)
2021-12-10 08:13:18.782,"  File ""/opt/latest/lib/python3.7/site-packages/services/metadata_service/api/artifact.py"", line 214, in get_artifacts_by_task"
2021-12-10 08:13:18.782,    artifacts.body)
2021-12-10 08:13:18.782,"  File ""/opt/latest/lib/python3.7/site-packages/services/data/db_utils.py"", line 70, in filter_artifacts_for_latest_attempt"
2021-12-10 08:13:18.782,    attempt_ids = get_latest_attempt_id_for_tasks(artifacts)
2021-12-10 08:13:18.782,"  File ""/opt/latest/lib/python3.7/site-packages/services/data/db_utils.py"", line 65, in get_latest_attempt_id_for_tasks"
2021-12-10 08:13:18.782,"    artifact['attempt_id'], attempt_ids.get(artifact['task_id'], 0))"
2021-12-10 08:13:18.782,TypeError: string indices must be integers
2021-12-10 08:13:18.782,"INFO:aiohttp.access:10.202.1.122 [10/Dec/2021:08:12:18 +0000] ""GET /flows/MRPFlow/runs/sfn-asdfasdfasfasdfadsfasdfdas/steps/some_step/tasks/asdfasfasdfasfsafasfasdgfasdgasf/artifacts HTTP/1.1"" 500 244 ""-"" ""python-requests/2.25.1"""
```

## What Is the Root Cause
We are not handing the exception from the database cleanly over [here](https://github.com/Netflix/metaflow-service/blob/65e19aef268e9e707522ee0695fd4ebaee42aa69/services/metadata_service/api/artifact.py#L209); In that line `artifacts` is a `DBResponse` object which consists of a status code. IN this [line](https://github.com/Netflix/metaflow-service/blob/65e19aef268e9e707522ee0695fd4ebaee42aa69/services/metadata_service/api/artifact.py#L209) we don't validate the response code of the DB's response and directly pass the `artifacts` to `filter_artifacts_for_latest_attempt`. This is where the service is failing. 

Based on the logs shared, we are getting a log like the following `2021-12-10 08:12:10.755,ERROR:AsyncPostgresDB:global:Exception occured `; This is emitted by [aiopg_exception_handling](https://github.com/Netflix/metaflow-service/blob/65e19aef268e9e707522ee0695fd4ebaee42aa69/services/data/db_utils.py#L13) function. Ideally, we should be checking the status_code and then making the call to `filter_artifacts_for_latest_attempt`

This PR tries to gracefully Handle such errors thrown by the Database.

### DIFF
--- a/services/data/db_utils.py
+++ b/services/data/db_utils.py
@@ -67,11 +67,28 @@ def get_latest_attempt_id_for_tasks(artifacts):
 
 
 def filter_artifacts_for_latest_attempt(artifacts):
+    """
+    filters artifacts of latest task attempt. 
+    Args:
+        artifacts (List[ArtifactRow]): a dictionary form of `ArtifactRow`
+
+    Returns:
+        `list` of filtered artifacts for the latest attempt
+    """
     attempt_ids = get_latest_attempt_id_for_tasks(artifacts)
     return filter_artifacts_by_attempt_id_for_tasks(artifacts, attempt_ids)
 
 
 def filter_artifacts_by_attempt_id_for_tasks(artifacts, attempt_for_tasks):
+    """
+    filters artifacts based on the attempt_ids. 
+    Args:
+        artifacts (List[ArtifactRow]): a dictionary form of `ArtifactRow`
+        attempt_for_tasks ([dict]): dictionary of {task_id:attempt_id}
+
+    Returns:
+        `list` of artifacts filtered based on `attempt_id`
+    """
     result = []
     for artifact in artifacts:
         if artifact['attempt_id'] == attempt_for_tasks[artifact['task_id']]:
--- a/services/data/db_utils.py
+++ b/services/data/db_utils.py
@@ -1,8 +1,11 @@
+from typing import List
 import psycopg2
 import collections
 import datetime
 import time
 import json
+
+from .models import ArtifactRow
 
 DBResponse = collections.namedtuple("DBResponse", "response_code body")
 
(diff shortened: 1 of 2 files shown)


services/metadata_service/api/artifact.py
```diff
@@ -314,12 +322,16 @@ async def get_artifacts_by_step(self, request):
...
-        )
+        if artifacts.response_code == 200:
+          filtered_body = filter_artifacts_for_latest_attempt(
+              artifacts.body)
+          return web.Response(
+              status=artifacts.response_code, body=json.dumps(filtered_body)
+          )
+        else:
+          return web.Response(
+            status=500,body=json.dumps(http_500(artifacts.body))
```
oavdeev, line 333: should it pass through `artifacts.response_code` here?

services/metadata_service/api/artifact.py
```diff
@@ -314,12 +322,16 @@ async def get_artifacts_by_step(self, request):
...
             flow_name, run_number, step_name
         )
-
-        filtered_body = filter_artifacts_for_latest_attempt(
-            artifacts.body)
-        return web.Response(
-            status=artifacts.response_code, body=json.dumps(filtered_body)
-        )
+        if artifacts.response_code == 200:
+          filtered_body = filter_artifacts_for_latest_attempt(
```
oavdeev, line 326: while we're at it, maybe add a type signature  for `filter_artifacts_for_latest_attempt()` for readability, for the next person who'll try to figure out what's going on

oavdeev (changes requested):

services/metadata_service/api/artifact.py
valayDave: Done.

services/metadata_service/api/artifact.py
valayDave: I have added docstrings for those functions. Those can help understand the types for readability.

services/data/db_utils.py
```diff
@@ -67,11 +67,28 @@ def get_latest_attempt_id_for_tasks(artifacts):
...
 def filter_artifacts_by_attempt_id_for_tasks(artifacts, attempt_for_tasks):
+    """
+    filters artifacts based on the attempt_ids. 
+    Args:
+        artifacts (List[ArtifactRow]): a dictionary form of `ArtifactRow`
+        attempt_for_tasks ([dict]): dictionary of {task_id:attempt_id}
+
+    Returns:
+        `list` of artifacts filtered based on `attempt_id`
+    """
```
oavdeev, line 91: Eh you could as well use type annotations, they are already used pretty widely in `metaflow-service` codebase. And I don't think this style of docstring is used anywhere else here.

oavdeev (approved):

services/data/db_utils.py
```diff
@@ -66,29 +69,12 @@ def get_latest_attempt_id_for_tasks(artifacts):
...
-def filter_artifacts_for_latest_attempt(artifacts):
-    """
-    filters artifacts of latest task attempt. 
-    Args:
-        artifacts (List[ArtifactRow]): a dictionary form of `ArtifactRow`
-
-    Returns:
-        `list` of filtered artifacts for the latest attempt
-    """
+def filter_artifacts_for_latest_attempt(artifacts:List[ArtifactRow]) -> List[ArtifactRow]:
```
valayDave, line 72: 🙈 ; I have changed to type hints but they are not the most accurate. This change has some nuance; `List[ArtifactRow]` is, in fact, list having a dictionary representation `ArtifactRow`

PR #273, opened 2021-12-18
feat: ui_backend support for cards

saikonen:
add support for getting a list of cards for a task, and the specific cards html content.
Routes added:
```
/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/cards
/flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks/{task_id}/cards/{hash}
```

TODO:

- [x] Remove custom pip install from dockerfile once card support is released in metaflow client.
- [x] Update metaflow client version in requirements.
- [x] change `METAFLOW_S3_RETRY_COUNT` back to 0 once metaflow client handles it correctly.
- [x] add integration tests to cover card api behaviour.


Dockerfile.ui_service
```diff
@@ -28,5 +28,6 @@ WORKDIR /root
 RUN /root/services/ui_backend_service/download_ui.sh
 
 RUN pip install --editable .
+RUN pip install git+https://github.com/valayDave/metaflow.git --upgrade
```
saikonen, line 31: remove this once testing is done and card support is completely released with metaflow client.

services/ui_backend_service/data/cache/get_cards_action.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        def _card_item(card):
+            return {
+                "type": card.type,
+                "html": card.get()
+            }
+        try:
+            with streamed_errors(stream_output):
+                task = Task("{}".format(pathspec))
+                cards = {index: _card_item(card) for index, card in enumerate(get_cards(task))}
```
saikonen, line 57: change index to `card.hash` once the client exposes this.

services/ui_backend_service/ui_server.py
```diff
@@ -37,7 +37,8 @@
 # Provide defaults for Metaflow Client
 os.environ['METAFLOW_SERVICE_URL'] = os.environ.get('METAFLOW_SERVICE_URL', DEFAULT_METADATA_SERVICE_URL)
 os.environ['USERNAME'] = os.environ.get('USERNAME', 'none')
-os.environ['METAFLOW_S3_RETRY_COUNT'] = os.environ.get('METAFLOW_S3_RETRY_COUNT', '0')
+# TODO: drop this back to 0 once the client is fixed to handle it correctly.
+os.environ['METAFLOW_S3_RETRY_COUNT'] = os.environ.get('METAFLOW_S3_RETRY_COUNT', '1')
```
saikonen, line 41: note the todo.

Dockerfile.ui_service
```diff
@@ -28,5 +28,6 @@ WORKDIR /root
 RUN /root/services/ui_backend_service/download_ui.sh
 
 RUN pip install --editable .
+RUN pip install git+https://github.com/saikonen/metaflow.git@card-client --upgrade
```
saikonen, line 31: remove this once testing is done and card support is completely released with metaflow client.

services/ui_backend_service/api/card.py
```diff
@@ -0,0 +1,207 @@
...
+    if pathspec in data:
+        success, value, detail, trace = unpack_processed_value(data[pathspec])
+        if success:
+            return value
+    # TODO: handle persisted errors with card content processing?
+
+    return None
+
+
+def get_pagination_params(request):
```
oavdeev, line 184: It seems like there is another, slightly different implementation of `get_pagination_params()` elsewhere in the codebase, in `log.py`, and another in [utils.py](https://github.com/Netflix/metaflow-service/blob/master/services/ui_backend_service/api/utils.py#L74:L74), maybe worth consolidating this logic.

Alternatively we could just remove pagination from this call altogether for now, as the list of tuples shouldn't get too big, and all filtering here is done on the python side anyways. UI backend will OOM anyways if the list gets too big. We can always add that later.

services/ui_backend_service/api/card.py
```diff
@@ -0,0 +1,207 @@
...
+            if event["type"] == "error":
+                # raise error, there was an exception during fetching.
+                raise CardException(event["message"], event["id"], event["traceback"])
+        await res.wait()  # wait until results are ready
+    data = res.get()
+    if pathspec in data:
+        success, value, detail, trace = unpack_processed_value(data[pathspec])
+        if success:
+            return value
+    # TODO: handle persisted errors with card content processing?
```
oavdeev, line 179: Can you elaborate what are these "persisted errors"? who persist them and where?

Also, is this a TODO item for this PR?

services/ui_backend_service/api/card.py
```diff
@@ -0,0 +1,207 @@
...
+          - $ref: '#/definitions/Params/Custom/invalidate'
+          - $ref: '#/definitions/Params/Custom/user_name'
+          - $ref: '#/definitions/Params/Custom/ts_epoch'
+        produces:
+        - application/json
+        responses:
+            "200":
+                description: Returns all artifacts of specified task
+                schema:
+                  $ref: '#/definitions/ResponsesArtifactList'
```
oavdeev, line 139: Looks like this needs to be changed too

oavdeev (changes requested):

oavdeev:
I think another todo item would be to add integration tests for these new APIs

services/ui_backend_service/api/card.py
```diff
@@ -0,0 +1,207 @@
...
```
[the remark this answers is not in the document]
saikonen: renamed some of the functions to be more descriptive, and updated apidocs related to card routes.

the return from this list route is mainly so the UI has some way to check which cards are fetchable, so the minimum requirement is to provide unique ids for fetching a single cards HTML content (with the card hash). We can further add fields if necessary, and this is mainly limited to what the `Card` class from the Metaflow client library provides.

services/ui_backend_service/api/card.py
saikonen: yes the todo was valid for this PR as well. This relates to the cache in general, and is perhaps not applicable to card content as there is no parsing of the values etc. associated, but for any exception that is deemed "unrecoverable" for the backend, we choose to persist the exception in the cache as a value, to skip further executions. 

One example would be calling an action for fetching S3 content with an invalid S3 location.

services/ui_backend_service/api/card.py
saikonen: The reason I included some pagination here is mainly to keep api list responses consistent, as the body includes pagination params, so there would perhaps be the expectation that they would affect the results as well.

Not sure whether consolidating the pagination helper from all three will clean anything up, as the helper for logs/cards only strips some query params or sets defaults but the utils helper is more closely tied to the sql query construction. I'll give it some further thought though.

oavdeev (approved):
lgtm aside from the TODO item left

PR #274, opened 2021-12-20
fix: add migration for missing string format indices.

saikonen:
Running a flow on SFN creates string format ids for `run_id` and `task_name`. Using these ids with metaflow client to access records is missing the necessary indices, and can result in very slow requests.

TODO:
- [ ]  Verify that the addition of the indices has no adverse effects on metadata service performance


romain-intel (approved):
LGTM. I think I have similar indices internally already.

oavdeev:
LGTM but looks like the TODO item still needs to be done before merging?

saikonen:
> LGTM but looks like the TODO item still needs to be done before merging?

If @romain-intel can confirm that they are indeed running the same indices internally then I'd consider these tested and ready for merging.

PR #280, opened 2022-01-03
Fixing error message responses for `psycopg2.Error`

valayDave:
Metaflow service now fails with error messages but `psycopg2.Error` objects are not playing well with string serialization `str(exception)` . Adding a fix to make the error messages more understandable.


services/data/postgres_async_db.py
```diff
@@ -242,6 +242,7 @@ async def execute_sql(self, select_sql: str, values=[], fetch_single=False,
             return aiopg_exception_handling(error), None
         except (Exception, psycopg2.DatabaseError) as error:
             self.db.logger.exception("Exception occured")
+            print(error)
```
saikonen, line 245: is this some leftover from testing? Otherwise if you need to print out the exception, the `logger.exception()` should already be taking care of this automatically.

saikonen (approved):
Looks like an excellent addition. Remember to clean up before merging

services/data/postgres_async_db.py
valayDave: Thank you for posting this anyways. This is a good reminder as I may have forgotten. I have removed this :)

services/data/db_utils.py
```diff
@@ -14,6 +15,28 @@
 def aiopg_exception_handling(exception):
     err_msg = str(exception)
     body = {"err_msg": err_msg}
+    if isinstance(exception, asyncio.TimeoutError):
+        body = {
+            "err_msg": {
+                "type": "timeout error",
+            }
+        }
+    elif psycopg2.Error in type(exception).__bases__:
```
oavdeev, line 24: what's the difference between this and `isinstance(exception, psycopg2.Error)` ?

valayDave: None actually. I can change that.

oavdeev (approved):

PR #281, opened 2022-01-05
Feat/api support for cards

obgibson:


saikonen:
Is this PR for testing purposes only or meant to be merged on top of #273 ?

obgibson:
> Is this PR for testing purposes only or meant to be merged to #273 ?

This is not just for testing purposes. It should be merged down. It was created so that we could have different versions of `metaflow` when deploying.

PR #282, opened 2022-01-12
docs: add local metadata service usage instructions to readme.

saikonen:
Add instructions on how to execute flows against a local metadata service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,17 @@ The test suite requires a PostgreSQL database, along with the following environm
 > tox -e unit,integration -p
 > ```
 
+### Executing flows against a local Metadata service
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,6 +130,16 @@ The test suite requires a PostgreSQL database, along with the following environm
 > tox -e unit,integration -p
 > ```
 
+### Executing flows against a local Metadata service
+
+With the metadata service up and running at `http://localhost:8080`, you are able to use this as the service when executing Flows with the Metaflow client locally via
+
+```sh
+METAFLOW_SERVICE_URL=http://localhost:8080 python3 basicflow.py run
```
rsanteri, line 138: Do we also need to include `--metadata=service` parameter here?

saikonen: good catch, added it as well.

PR #283, opened 2022-01-18
Extra logging for JSON errors

obgibson:
I have had a lot of problems with setting the `PLUGINS` environment variable, which requires a JSON string as its value.

This PR adds some extra logging to the JSON decoding to aid in debugging when the supplied JSON is poorly formed.


services/ui_backend_service/api/utils.py
```diff
@@ -11,12 +11,17 @@
...
+
+logger = logging.getLogger("Utils")
 
 
 def get_json_from_env(variable_name: str):
     try:
         return json.loads(os.environ.get(variable_name))
-    except Exception:
+    except Exception as e:
+        logger.warning("error parsing JSON: %s, from %s: %s", e, variable_name, os.environ.get(variable_name))
```
saikonen, line 24: Excellent info with the added logging. The old style string interpolation is mainly used for passing string templates around on the service side, so for consistency the warning formatting could instead use .format() or an f-string, whichever is more readable? Up to you if the change makes sense to implement :)

saikonen (approved):

services/ui_backend_service/api/utils.py
saikonen: consider specifically catching json parsing errors instead though, as we usually _try_ to load configs from env, but most of the time it is not an exception if an env_var is not set, so we probably do not want a warning in these cases.

saikonen: consider specifically catching json parsing errors instead though, as we usually try to load configs from env, but most of the time it is not an exception if an env_var is not set, so we probably do not want a warning in these cases.

saikonen (changes requested):

saikonen (approved):

saikonen:
changes look good, approved :) grab the changed dev dependencies from `master` in order to get the integration tests to pass for this as well.

PR #292, opened 2022-01-27
Fix: cache issues

saikonen:
- Adds a check that tmpdir exists when launching workers, and creates the dir if it is missing. This fixes workers not being able to launch if for any reason the cache directory gets wiped.
- Adds log cache process to the cache healthcheck monitoring, so it will correctly be restarted in case of issues.


ruial:
This fix is very much needed, the LogCacheStore tends to die after a while and the service requires a manual restart.

Could someone please review? @oavdeev @obgibson @jackie-ob

valayDave (approved):

services/ui_backend_service/data/cache/store.py
```diff
@@ -77,7 +77,7 @@ async def start_caches(self, app):
 
     async def _monitor_restart_requests(self):
         while True:
-            for _cache in [self.artifact_cache, self.dag_cache]:
```
valayDave, line 80: We should merge this change.

Issue #311, opened 2022-04-13
Building with `Dockerfile.ui_service` fails due to old version of pip


hallsop:
The following error is received when attempting to build the `Dockerfile.ui_service` file using `docker-compose.development.yml`

```
ERROR: Could not build wheels for pygit2 which use PEP 517 and cannot be installed directly
WARNING: You are using pip version 19.1.1, however version 22.0.4 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
The command '/bin/sh -c pip install --editable .' returned a non-zero code: 1
ERROR: Service 'ui_backend' failed to build
```

The version of pip needs to be upgraded in the Dockerfile to fix this issue

PR #312, opened 2022-04-13
Upgrade pip in `Dockerfile.ui_service` to version 22.0.4

hallsop:
Fix https://github.com/Netflix/metaflow-service/issues/311 by upgrading pip in `Dockerfile.ui_service`

### DIFF
--- a/Dockerfile.ui_service
+++ b/Dockerfile.ui_service
@@ -27,7 +27,7 @@ WORKDIR /root
 # Install Netflix/metaflow-ui release artifact
 RUN /root/services/ui_backend_service/download_ui.sh
 
-RUN python3 -m pip install -- upgrade pip==22.0.4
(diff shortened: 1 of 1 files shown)


oavdeev:
Thanks for the PR! I failed to reproduce this but it seems like the culprit is that the `python:3.7` image you had locally had an older version of pip than the most recent one. If you do `docker pull python:3.7 && docker run -it --rm python:3.7 pip -V` today you'll see that it is already 22.0.4

I found that it has been upgraded in this commit https://github.com/docker-library/python/commit/a9815ead3ae016edf7039285cb267b14a5700169

I'm thinking maybe a slightly cleaner fix for this issue would be to pin _image version_ to 3.7.13 in `FROM` but continue using pip that comes with the base image.

hallsop:
Ah thanks! Yes that makes a lot more sense, have updated the PR

PR #315, opened 2022-05-09
[tag-mutation-project] Metadata service - object read paths to return ancestral Run tags

jackie-ob:
This is part of project to add tag mutation support.  RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Consolidate tagging to Runs")

In this PR, we change object read APIs on the metadata service for these object types (all of these belong to a specific Run):
* Step
* Task
* Artifact

In particular, all `tags` and `system_tags` fields of the above objects will be replaced with the same fields from their ancestral Run objects.  E.g. if a Run R has tasks T1, T2, T3, then the API will return T1 with its task level tags (as stored in DB) with R.tags.

Metadata was explicitly excluded even though it is a descendant of Run.  Each Task can experience multiple attempts, and each attempt gives us a fresh set of metadata records.  We use the `tags` field on these metadata records themselves to isolate attempt numbers!

As a drive-by, some flakiness in the metadata service integration test suite was fixed.  We used to test for specific return order of list results - which does not appear to be explicitly guaranteed by the backend - the tests have been updated to no longer depend on this ordering.


pytest.ini
```diff
@@ -1,4 +1,5 @@
 [pytest]
 markers =
     unit_tests: Unit tests (deselect with '-m "not unit_tests"')
-    integration_tests: Integration tests (deselect with '-m "not integration_tests"')
\ No newline at end of file
+    integration_tests: Integration tests (deselect with '-m "not integration_tests"')
+asyncio_mode=auto
```
jackie-ob, line 5: pytest warned me about the default being "legacy" asyncio mode.  Recommendation is to at least set to `auto` (and underneath it resolves to legacy).

services/metadata_service/api/artifact.py
```diff
@@ -59,6 +60,7 @@ def __init__(self, app):
             self.create_artifacts,
         )
         self._async_table = AsyncPostgresDB.get_instance().artifact_table_postgres
+        self._async_run_table = AsyncPostgresDB.get_instance().run_table_postgres
```
jackie-ob, line 63: This will be a common pattern.  Various API modules now require the run table.  Not a new pattern.  `TaskApi` already did this.

services/metadata_service/api/artifact.py
```diff
@@ -103,15 +105,17 @@ async def get_artifact(self, request):
             "405":
                 description: invalid HTTP Method
         """
-        flow_name = request.match_info.get("flow_id")
+        flow_id = request.match_info.get("flow_id")
```
jackie-ob, line 108: `flow_name` seems wrong based on current code (though I'm sure there is a history behind this).  So I fixed the var names in the code I touched.

services/metadata_service/tests/integration_tests/artifact_test.py
```diff
@@ -132,8 +131,12 @@ async def test_run_artifacts_get(cli, db):
     _first_artifact = (await add_artifact(db, flow_id=_task["flow_id"], run_number=_task["run_number"], step_name=_task["step_name"], task_id=_task["task_id"], artifact=ARTIFACT_A)).body
     _second_artifact = (await add_artifact(db, flow_id=_task["flow_id"], run_number=_task["run_number"], step_name=_task["step_name"], task_id=_task["task_id"], artifact=ARTIFACT_B)).body
 
+    # expect artifacts' tags to be overridden by tags of their ancestral run
+    update_objects_with_run_tags('artifact', [_first_artifact, _second_artifact], _run)
+
     # try to get all the created artifacts
-    await assert_api_get_response(cli, "/flows/{flow_id}/runs/{run_number}/artifacts".format(**_task), data=[_first_artifact, _second_artifact])
+    await assert_api_get_response(cli, "/flows/{flow_id}/runs/{run_number}/artifacts".format(**_task),
+                                  data=[_first_artifact, _second_artifact], data_is_unordered_list=True)
```
jackie-ob, line 139: When I ran the integration test suite locally, there was flakiness whenever we depend on returned object order (roughly fails 10-30% of the time as a suite).  We stop enforcing result ordering in all the places for `assert_api_get_response`.  New callers will still default to enforcing ordering to not surprise anyone here.

services/metadata_service/tests/integration_tests/utils.py
```diff
@@ -209,17 +220,32 @@ async def assert_api_get_response(cli, path: str, status: int = 200, data: objec
...
 
     assert response.status == status
 
-    if data:
-        body = json.loads(await response.text())
+    body = json.loads(await response.text())
+    if data is None:
+        return
+    if data_is_unordered_list:
+        assert type(data) == list and type(body) == list
```
oavdeev, line 234:
```suggestion
        assert isinstance(data, list) and isinstance(body, list)
```
I think `isinstance` is a lil more idiomatic and more proof if the value is some kind of a list derivative.

services/metadata_service/tests/integration_tests/utils.py
```diff
@@ -209,17 +220,32 @@ async def assert_api_get_response(cli, path: str, status: int = 200, data: objec
         http status code to expect from response
     data : object
         An object to assert the api response against.
+    data_is_unordered_list : bool
+        Data is an unordered list, so ignore ordering when comparing data and response body
```
oavdeev, line 224: seems like it assumes that data is an unordered list _of dictionaries_ , right?

services/metadata_service/tests/integration_tests/utils.py
```diff
@@ -209,17 +220,32 @@ async def assert_api_get_response(cli, path: str, status: int = 200, data: objec
...
-    if data:
-        body = json.loads(await response.text())
+    body = json.loads(await response.text())
+    if data is None:
+        return
+    if data_is_unordered_list:
+        assert type(data) == list and type(body) == list
+        if not data:
+            assert body == []
+            return
```
oavdeev, line 237: does it need a special case here? i thought the assertion a few lines below would still handle that?

oavdeev (commented):
overall lgtm 👍  except for some minor comments

services/metadata_service/api/tagging_utils.py
```diff
@@ -0,0 +1,25 @@
+from services.data.db_utils import DBResponse
+
+
+async def replace_with_run_tags_in_db_response(flow_id, run_number, run_table_postgres, db_response: DBResponse) -> DBResponse:
+    # Only replace tags if response code is legit
+    # Object creation ought to return 201 (let's prepare for that)
```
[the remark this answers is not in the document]
jackie-ob: Names are very important, and is the hardest thing in computer science.  I picked a slightly better name I think and added the comment.

I also changed the code to return a copy of the modified db response instead - I think that is probably less surprising / more Pythonic.  We can afford it.

services/metadata_service/tests/integration_tests/utils.py
jackie-ob: You are right, will remove that.  Some previous version was looking for the first element to detect fields (not needed after all).

oavdeev:
One additional thing that came to mind: maybe worth adding some comments on the _write path_ for step/task etc. mentioning that the tags are now going to a black hole?

jackie-ob:
> 

Good idea, done.

romain-intel (approved):
LGTM. There seems to be a tiny change in behavior for heartbeat on task creation (if for some reason we get a 500 on getting the run but not on creating the task we won't update hb) but I don't think that is ever a thing that could happen. O/w looks fine to me.

oavdeev (approved):

PR #317, opened 2022-05-17
[tag-mutation-project] UI backend service - object read paths to return ancestral Run tags

jackie-ob:
This is part of project to add tag mutation support. RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Consolidate tagging to Runs")

This PR can be considered fully analogous to this [PR for the Metadata service](https://github.com/Netflix/metaflow-service/pull/315) (same repo, different services).

We moved `services/metadata_service/api/tagging_utils.py` to a shared location `services/data/` - it turns out we can still use it for UI backend code.  The two services nominally use different DB abstraction objects, but duck-typing says it's OK (both have very similar APIs).

UI backend service code has an existing concept of response "postprocessing".  We leverage this as appropriate to apply Run-level tags to Step, Task and Artifact object read paths.

Drive-by: Fixed some flaky tests (due to item ordering).


services/ui_backend_service/api/heartbeat_monitor.py
```diff
@@ -201,7 +203,7 @@ async def add_to_watch(self, data):
             if heartbeat_ts is not None:  # only start monitoring on runs that have a heartbeat
                 self.watched[key] = heartbeat_ts
 
-    async def get_task(self, flow_id: str, run_key: str, step_name: str, task_key: str, attempt_id: int = None, postprocess=None) -> Optional[Dict]:
+    async def _get_task(self, flow_id: str, run_key: str, step_name: str, task_key: str, attempt_id: int = None, postprocess=None) -> Optional[Dict]:
```
jackie-ob, line 206: Renamed this to make it pseudo-private.  It is used in a private manner only.

services/ui_backend_service/tests/integration_tests/metadata_test.py
```diff
@@ -138,18 +138,18 @@ async def test_list_metadata_attempt_id_filter(cli, db):
     _metadata_second["attempt_id"] = 1
 
     # no filter
-    await _test_list_resources(cli, db, "/flows/{flow_id}/runs/{run_number}/metadata?order=ts_epoch".format(**_task), 200, [_metadata_second, _metadata_first, _metadata_global])
+    await _test_list_resources(cli, db, "/flows/{flow_id}/runs/{run_number}/metadata?_order=ts_epoch".format(**_task), 200, [_metadata_second, _metadata_first, _metadata_global])
```
jackie-ob, line 141: `order` => `_order`.  It never worked because we only look for `_order` on backend...

services/ui_backend_service/tests/integration_tests/tasks_test.py
```diff
@@ -494,17 +497,32 @@ async def create_ok_artifact_for_task(db, task, attempt=0):
     return _task
 
 
-async def create_task(db, step=None, status="running", task_id=None, task_name=None, last_heartbeat_ts=None):
```
jackie-ob, line 497: I forked this.  Previously `step` is an optional arg.  If no `step` is provided, `create_task` created its own stack of flow, run, step because creating the task.  We did not return any of the created items, and it would have been awkward to do so (return a tuple?).

Instead I separated into `create_task_with_step` - where the caller is responsible for preparing ancestor objects.  It then becomes clearer exactly when and how we apply ancestral run tags on a created task.

services/ui_backend_service/tests/integration_tests/utils.py
```diff
@@ -275,12 +275,29 @@ async def _test_list_resources(cli, db: AsyncPostgresDB, path: str, expected_sta
...
+
+    data_to_be_compared = data
+    expected_data_to_be_compared = expected_data
+    if ignore_data_order:
+        assert isinstance(data, list) and isinstance(expected_data, list)
+
+        # If item contains fields A and B, then sort list first by item[A], then item[B]
+        # Exclude approx_keys, because we don't want approx_sort.
+        def _sort_key(r):
+            if isinstance(r, str):
```
jackie-ob, line 287: The data list can contain `str` items sometimes.  We handle it.  If other types show up in future we will need to add more rules - we won't miss it.

oavdeev (approved):

savingoyal (approved):

romain-intel (approved):
Didn't look in huge detail but just to let you know that nothing stands out :)

PR #320, opened 2022-05-25
[tag-mutation-project] Add tag mutation endpoint for Runs to metadata service

jackie-ob:
This is part of project to add tag mutation support. RFC [here](https://www.notion.so/outerbounds/RFC-Implementing-tag-mutation-98ef6ba39aee4aeeb9c6150b4cffea2a) (see "Tagging CLI").

The new Metaflow client API (which tagging CLI will use) will call this new metadata service endpoint:

```
        app.router.add_route("PATCH",
                             "/flows/{flow_id}/runs/{run_number}/tag/mutate",
```

Some requirements:

- A - Tag changes should be atomic (e.g. we can add/remove multiple tags at once).
- B - Tag changes should not be lost for any concurrent requests.
- C - Fail if mutation is invalid w.r.t. to current tag content (e.g. system tags)

We implement tag updates in the most straightforward manner. In DB terms:

- Check mutation request inputs (without needing current run state)
- Start DB transaction - serializable isolation level (B)
  - Read current run system tags
  - Check mutation request inputs vs current system tags (C)
  - Perform user tag mutation in memory
  - Persist new full list of user tags to DB (A)
- Commit DB transaction.

With B, concurrent mutation requests may fail with 503, if another mutation is inflight. It will be up to the Metaflow client API to retry. We will implement a sensible retry policy in API client so that end user of the client API will not need to care, except in extreme circumstances.

Run tags are stored as a JSON list in a PostgresDB JSONB column. Why not use native Postgres JSON operations?

Pros:

- Efficiency for large tag sets (e.g. if modifying one tag.  O(N) either way, but less DB write load)

Cons:

- JSON only has lists, not sets. We will need to reconcile this in memory (e.g. for removals, we would need to build an index to know which position to remove, etc).
- JSON only has lists, not sets. It means we still need to explicitly take care to prevent dupes ending up in DB somehow.
- A native JSON manipulation implementation will be less readable, less "clearly correct".

The decision in this PR is to go with simple. Efficiency-wise: there will be no issues unless folks send hundreds of tag mutation requests per second and persisted tag lists are long (hundreds of tags). We can revisit this implementation detail in future if/when needed without changing outward facing API.

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -162,7 +162,8 @@ class AsyncPostgresTable(object):
             await PostgresUtils.setup_trigger_notify(db=self.db, table_name=self.table_name, keys=self.trigger_keys)
 
     async def get_records(self, filter_dict={}, fetch_single=False,
-                          ordering: List[str] = None, limit: int = 0, expanded=False) -> DBResponse:
+                          ordering: List[str] = None, limit: int = 0, expanded=False,
+                          cur: aiopg.Cursor = None) -> DBResponse:
         conditions = []
         values = []
         for col_name, col_val in filter_dict.items():
@@ -171,13 +172,13 @@ class AsyncPostgresTable(object):
 
         response, _ = await self.find_records(
             conditions=conditions, values=values, fetch_single=fetch_single,
-            order=ordering, limit=limit, expanded=expanded
+            order=ordering, limit=limit, expanded=expanded, cur=cur
         )
         return response
 
     async def find_records(self, conditions: List[str] = None, values=[], fetch_single=False,
                            limit: int = 0, offset: int = 0, order: List[str] = None, expanded=False,
-                           enable_joins=False) -> Tuple[DBResponse, DBPagination]:
+                           enable_joins=False, cur: aiopg.Cursor = None) -> Tuple[DBResponse, DBPagination]:
         sql_template = """
         SELECT * FROM (
             SELECT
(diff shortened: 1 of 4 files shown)


services/data/postgres_async_db.py
```diff
@@ -162,7 +162,8 @@ async def _init(self, create_triggers: bool):
             await PostgresUtils.setup_trigger_notify(db=self.db, table_name=self.table_name, keys=self.trigger_keys)
 
     async def get_records(self, filter_dict={}, fetch_single=False,
-                          ordering: List[str] = None, limit: int = 0, expanded=False) -> DBResponse:
+                          ordering: List[str] = None, limit: int = 0, expanded=False,
+                          cur: aiopg.Cursor = None) -> DBResponse:
```
jackie-ob, line 166: Cursor lifecycle is closely tied to DB transaction lifecycle

For tag mutation, we want to use the same cursor when executing logic that is dispersed across multiple existing DB access functions (read run, do some checks, write run tag).  The idea is to pass in your own cursor as an optional in these existing DB access functions.  If we pass in the cursor, the inner DB code will use that (without managing its lifetime, like "closing" it), instead of always creating a new one.

services/data/postgres_async_db.py
```diff
@@ -286,10 +291,28 @@ async def create_record(self, record_dict):
...
+            return aiopg_exception_handling(error)
+
+    async def run_in_transaction_with_serializable_isolation_level(self, fun):
+        try:
+            with (
+                    await self.db.pool.cursor(
+                        cursor_factory=psycopg2.extras.DictCursor,
+                    )
+            ) as cur:
+                async with cur.begin():
```
jackie-ob, line 304: This is the one cursor / DB transaction wrapping most of our mutation operation.

services/data/postgres_async_db.py
```diff
@@ -510,6 +537,17 @@ async def update_heartbeat(self, flow_id: str, run_id: str):
         return DBResponse(response_code=result.response_code,
                           body=json.dumps(body))
 
+    async def update_run_tags(self, flow_id: str, run_id: str, run_tags: list, cur: aiopg.Cursor = None):
+        run_key, run_value = translate_run_key(run_id)
+        filter_dict = {"flow_id": flow_id,
+                       run_key: str(run_value)}
+        set_dict = {"tags": "'" + json.dumps(run_tags) + "'"}  # TODO we can/should do better than manual quoting
```
oavdeev, line 544: May be a good idea to sort them before storing. Not sure if it makes any practical difference but its nice from debuggability perspective that e.g. after removing and adding a tag you'd have exact same value in the db.

oavdeev, line 544: do you want to fix this TODO before merging?

services/metadata_service/api/run.py
```diff
@@ -136,6 +143,97 @@ async def create_run(self, request):
...
+
+        if not isinstance(tags_to_add, list):
+            return DBResponse(response_code=422, body="tags_to_add must be a list")
+
+        if not isinstance(tags_to_remove, list):
+            return DBResponse(response_code=422, body="tags_to_remove must be a list")
+
+        # let's make sure we have a list of strings
+        if not all(isinstance(t, str) for t in chain(tags_to_add, tags_to_remove)):
+            return DBResponse(response_code=422, body="All tag values must be strings")
```
oavdeev, line 200: May be a good idea to validate the string format while we're at it

oavdeev (commented):
left some suggestions but overall lgtm

services/data/postgres_async_db.py
jackie-ob: Discussed offline.  My view is to follow [this](https://www.hyrumslaw.com/) here.  I get the point about potentially improved debuggability.

My preference is to leave it unordered until we are ready to commit to promising this in the public contract of the API.

jackie-ob: Fixed

services/metadata_service/api/run.py
jackie-ob: We will take care of this in `update_run_tags` above with `QuotedString`

oavdeev (approved):

services/metadata_service/api/run.py
```diff
@@ -136,6 +143,97 @@ async def create_run(self, request):
...
+            type: object
+            properties:
+                tags_to_add:
+                    type: array of string
+                tags_to_remove:
+                    type: array of string
+        produces:
+        - 'text/plain'
+        responses:
+            "200":
```
romain-intel, line 179: nit: 503 response?

services/metadata_service/api/run.py
```diff
@@ -136,6 +143,97 @@ async def create_run(self, request):
...
+                description: invalid HTTP Method
+        """
+        flow_name = request.match_info.get("flow_id")
+        run_number = request.match_info.get("run_number")
+        body = await read_body(request.content)
+        tags_to_add = body.get("tags_to_add", [])
+        tags_to_remove = body.get("tags_to_remove", [])
+
+        if not isinstance(tags_to_add, list):
+            return DBResponse(response_code=422, body="tags_to_add must be a list")
```
romain-intel, line 193: It's debatable whether to return a 400 or a 422 here. No strong opinion but 422 is a bit more obscure than 400. As long as the client deals with it ok, I suppose it is fine though.

services/metadata_service/api/run.py
```diff
@@ -136,6 +143,97 @@ async def create_run(self, request):
...
+            if run_db_response.response_code != 200:
+                if run_db_response.response_code == 404:
+                    return run_db_response
+                return DBResponse(response_code=422,
+                                  body="Failed to get run (get_run status %d)" % run_db_response.response_code)
+            run = run_db_response.body
+            existing_tag_set = set(run["tags"])
+            existing_system_tag_set = set(run["system_tags"])
+
+            for tag in tags_to_remove_set:
```
romain-intel, line 216: same as for local modificationn; this can be a simpler set operation and return all errors at once instead of one at a time.

romain-intel (approved):
Minor comments but otherwise looks good.

services/metadata_service/api/run.py
jackie-ob: Thanks! Missing various new status codes here.  Fixing.

services/metadata_service/api/run.py
jackie-ob: Done!

PR #321, opened 2022-05-31
[tag-mutation-project] Tag mutation endpoint fix for double JSON encoding

jackie-ob:
The `/tag/mutate` is expected to return a JSON response:

```
{
   "tags": LIST_OF_LATEST_TAGS_AFTER_MUTATION
}
```

We erroneously serialized the JSON object one too many times on the way out.  This is the fix.

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -545,11 +545,9 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
                        run_key: str(run_value)}
 
         set_dict = {"tags": QuotedString(json.dumps(run_tags)).getquoted().decode()}
-        result = await self.update_row(filter_dict=filter_dict,
-                                       update_dict=set_dict,
-                                       cur=cur)
-        return DBResponse(response_code=result.response_code,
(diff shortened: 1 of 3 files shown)


services/data/postgres_async_db.py
```diff
@@ -545,11 +545,9 @@ async def update_run_tags(self, flow_id: str, run_id: str, run_tags: list, cur:
                        run_key: str(run_value)}
 
         set_dict = {"tags": QuotedString(json.dumps(run_tags)).getquoted().decode()}
-        result = await self.update_row(filter_dict=filter_dict,
-                                       update_dict=set_dict,
-                                       cur=cur)
-        return DBResponse(response_code=result.response_code,
-                          body=json.dumps(set_dict))
+        return await self.update_row(filter_dict=filter_dict,
```
jackie-ob, line 548: WEe will use the DBResponse.body from `update_run_tags` if there was a non-200 response.  So let's remove this extraneous JSON encode step in case we do see it.  Let's just return whatever `update_row` says.

oavdeev (approved):

PR #328, opened 2022-08-15
Improve CORS support in UI backend

jackie-ob:
In deployments where UI static and UI backend services have different origins, preflight CORS checks fail.  This is true if the deployment architecture does not include any HTTP level routing layer that pulls UI static + UI backend services under a single origin (e.g. with k8s ingress, or other ALB).  E.g. for local development, running on localhost.

See: https://share.getcloudapp.com/BluGrpKB

We add an env var `ORIGIN_TO_ALLOW_CORS_FROM` to UI static backend so allow such deployments to work correctly.  E.g. if UI backend shall be hosted on `http://localhost:3000` (with UI backend hosted elsewhere, like `http://localhost:8083`), then set this in UI backend service's environment:

`ORIGIN_TO_ALLOW_CORS_FROM="http://localhost:3000"`.

Note that it may be set to `*` as well, if permitting all origins is OK.


services/ui_backend_service/frontend.py
```diff
@@ -30,11 +30,11 @@ def __init__(self, app):
         for filepath in static_files:
             filename = filepath[len(static_ui_path) + 1:]
             app.router.add_route(
-                '*', f'/{filename}', self.serve_file(filename))
```
jackie-ob, line 33: Setting "*" method clashes with `aiohttp_cors`, because it overlaps with the `OPTIONS` method needed for preflight CORS handling.

I'm assuming that static stuff only needs `GET`.  When I tested this, it seems to work fine.  Would love to hear any tribal knowledge about the reasoning behind existing `*` catch all route.

romain-intel (approved):
LGTM

PR #341, opened 2023-01-18
Drop all partial indexes and replace them with full indexes

sarthak:
Context: [Please see this slack discussion](https://outerbounds-community.slack.com/archives/C020U025QJK/p1673868236239179).

In all our tables we had some "partial" indexes such as this one:
```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS metadata_v3_idx_str_ids_primary_key ON metadata_v3 (
  id,
  flow_id,
  run_id,
  step_name,
  task_name,
  field_name
)
WHERE
  run_id IS NOT NULL
  AND task_name IS NOT NULL;
```
A "partial" index is one whose domain is constrained by a `WHERE` clause.
Such partial indexes can lead to sub-optimal query plans in unexpected ways.
For example, imagine a query like -
```sql
SELECT *
FROM metadata_v3
WHERE flow_id = ?
    AND run_id = ?
    AND step_name = ?
```
While we would like the query planner to choose an index scan, we will find
that query planner chooses to perform a sequential scan instead. The reason
is that the query doesn't constrain `task_name` to be non-null (and neither
does the table schema). Thus, from query planner's perspective the index is
effectively useless because it will not yield the full search space. These claims
are not theoretical or hypothetical, but verified. And such queries are not
artificial either but real queries constructed and executed by metaflow's
metadata and ui-backend service. At our scale (~10M rows in artifact_v3 table),
this directly led to extremely slow response times from metaflow dashboard.

With this PR, we propose to eliminate all partial indexes and replace
them with complete indexes instead. Although I would have liked to keep
everything else the same, along the way I discovered that certain indexes
specifically the one showed above incorrectly indexes the `id` column as
well (which is essentially a pkey column) rendering the whole index useless.
For such indexes we slightly edit them during re-creation to remove such
inefficiencies.

Caveat: For very large-scale metaflow deployments this migration might
be expensive (at our scale each index creation takes anywhere between
1-4 minutes depending upon load at the DB). I am worried that this might
cause our migrator lambda function to timeout. Looking for community's
opinion on this.

## Test Plan
Locally I tested that migration works by doing
```
docker-compose -f docker-compose.development.yaml up migration
```
and verifying that it is able to successfully apply the migration.
In my organisation's metaflow deployment we've also applied this
migration in production with no observed issues so far.


obgibson:
Tested the migration on a fresh instance. Confirmed that it worked.

Tested the migration on an older service and it failed - likely not related to this PR.

savingoyal (approved):

services/migration_service/migration_files/20230118020300_drop_partial_indexes.sql
```diff
@@ -0,0 +1,108 @@
...
+  ON runs_v3 (flow_id, run_id);
+
+DROP INDEX CONCURRENTLY IF EXISTS runs_v3_idx_str_ids_primary_key;
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS steps_v3_idx_str_ids_primary_key_v2
+  ON steps_v3 (flow_id, run_id, step_name);
+
+DROP INDEX CONCURRENTLY IF EXISTS steps_v3_idx_str_ids_primary_key;
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS tasks_v3_idx_flow_id_run_id_step_name_task_name
```
tmwong2003, line 18: This index has the same name as the one with the `WHERE ...` clauses. At best it doesn't create an index because of the `IF NOT EXISTS`; at worst it creates the index only for the next `DROP INDEX` to destroy it.
```suggestion
CREATE INDEX CONCURRENTLY IF NOT EXISTS tasks_v3_idx_flow_id_run_id_step_name_task_name_v2
```

services/migration_service/migration_files/20230118020300_drop_partial_indexes.sql
```diff
@@ -0,0 +1,108 @@
...
+  attempt_id,
+  name
+);
+
+DROP INDEX CONCURRENTLY IF EXISTS artifact_v3_idx_str_ids_primary_key;
+
+
+-- +goose Down
+
+-- copy of 20211202100726_add_str_id_indices.sql
```
tmwong2003, line 66: Should everything after this line really be here? It looks like the commands will undo everything from above the line.

tmwong2003 (commented):
@romain-intel @savingoyal I [created the indices as suggested](https://23andme.slack.com/archives/C02116BBNTU/p1685503411073869) to replace the indices with `WHERE ... NOT NULL` indices, and our query performance (and thus UI responsiveness) is orders of magnitude better.

services/migration_service/migration_files/20230118020300_drop_partial_indexes.sql
sarthak: This is a very good catch, yes it's a bug, and your suggestion is correct. I don't know how I missed this!

services/migration_service/migration_files/20230118020300_drop_partial_indexes.sql
```diff
@@ -0,0 +1,115 @@
...
+CREATE INDEX CONCURRENTLY IF NOT EXISTS artifact_v3_idx_str_ids_primary_key_v2 ON artifact_v3 (
+  flow_id,
+  run_id,
+  step_name,
+  task_name,
+  attempt_id,
+  name
+);
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS artifact_v3_idx_str_ids_primary_key_with_task_id ON artifact_v3 (
```
saikonen, line 53: same question for this index, is there a real use-case that requires it? Nothing is preventing a user from calling the API with a run_id and task_id of course, but we usually try to expose only one form of id when available.

saikonen:
With the feedback from testing, I'm leaning heavily towards solving this issue with the proposed full indices as well.

Sidenotes:
I had a quick stab at an alternative solution to utilizing the existing partial indices to see if a smaller fix would suffice (https://github.com/Netflix/metaflow-service/pull/363). By appending the constraint to queries where necessary, effectively nudging the query planner to use the correct index this way. There are some big drawbacks to this:
- changes are non-trivial and introduce a lot of unnecessary boilerplate. 
- affected areas span both metadata and ui services.
- forgetting to add the partian index constraints to any future query will lead to the same issue.

saikonen (approved):

PR #345, opened 2023-02-23
[UI backend] New options to restrict cards / logs loading

jackie-ob:
Introduces three new environment variables to influence cards and logs loading behavior:

`MF_LOG_LOAD_POLICY`
- `full` (default): loads entire log, matches existing behavior.
- `tail`: loads the tail of the log only (up to fixed size by number of characters).
- `blurb_only`: does not load the log at all. Instead return Python code snippet to access logs using Metaflow client.

`MF_LOG_LOAD_TAIL_SIZE`
Defaults to 100*1024.  Applies when `MF_LOG_LOAD_POLICY=tail`.  Returns the last N lines of the log file, where N is maximized without returning more than `MF_LOG_LOAD_TAIL_SIZE` number of characters.

`MF_CARD_LOAD_POLICY`
* `full` (default): Loads all cards, matches existing behavior.
* `blurb_only`: does not load cards at all.  Instead return Python code snippet to access cards using Metaflow client.

### DIFF
--- a/services/ui_backend_service/data/cache/get_log_file_action.py
+++ b/services/ui_backend_service/data/cache/get_log_file_action.py
@@ -81,11 +81,11 @@ class GetLogFile(CacheAction):
         result_key = log_result_id(task, logtype, limit, page, reverse_order, raw_log)
         stream_key = 'log:stream:%s' % lookup_id(task, logtype, limit, page, reverse_order, raw_log)
 
-        return msg, \
-               [log_key, result_key], \
-               stream_key, \
-               [stream_key, result_key], \
-               invalidate_cache
+        return (msg,
+                [log_key, result_key],
+                stream_key,
+                [stream_key, result_key],
+                invalidate_cache)
 
     @classmethod
     def response(cls, keys_objs):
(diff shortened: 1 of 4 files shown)


jackie-ob:
Want to get this out to gather feedback quickly.

I will update this PR with:
* In-repo doc updates
* Test coverage

jackie-ob:
Log tail mode:
image

jackie-ob:
Cards: blurb only
image

jackie-ob:
Logs: blurb only

image

savingoyal (approved):

services/ui_backend_service/data/cache/get_log_file_action.py
```diff
@@ -178,7 +199,87 @@ def get_log_content(task: Task, logtype: str):
...
-def paginated_result(content: List[Tuple[Optional[int], str]], page: int = 1, limit: int = 0, reverse_order: bool = False, output_raw=False):
+class LogProviderBase:
+
+    @staticmethod
+    def get_log_hash(task: Task, logtype: str) -> int:
+        raise NotImplementedError
+
+    @staticmethod
+    def get_log_content(task: Task, logtype: str):
+        raise NotImplementedError
```
oavdeev, line 210: You could make it `typing.Protocol` which would be arguably a little more elegant, and have get_log_provider return type be protocol type but up to you, don't feel strongly about this.

oavdeev (approved):

services/ui_backend_service/data/cache/get_log_file_action.py
jackie-ob: OK, I will add the tests and then try it out.

PR #346, opened 2023-03-06
CUS-42: Change cutoff time from 6 to 1 day

obgibson:
This PR changes the default `OLD_RUN_FAILURE_CUTOFF_TIME` and `RUN_INACTIVE_CUTOFF_TIME` values from 6 days to 1 day.

This change is so that tasks are not showing as still around when they have failed days ago.

The documentation for these variables has been updated.


obgibson:
Note: the majority of line changes are due to automatic formatting

Issue #352, opened 2023-03-13
Move on from Python 3.7 before security updates are discontinued. 


saikonen:
The main way that the services are released is through the published docker image. This, along with all of the service codebase is currently running and tested on Python 3.7, which is scheduled for EOL the coming summer (https://devguide.python.org/versions/).

As we do not have any backward compatibility requirements for the execution of the codebase due to the way it is released, it could be considered that we move to a more recent version of Python, along with updating the base image for the docker build.

dhpollack:
How many times can I upvote this?

saikonen:
Addressed in release [v2.4.2](https://github.com/Netflix/metaflow-service/releases/tag/v2.4.2)

PR #367, opened 2023-06-22
chore: update python version

saikonen:
In preparation for Python 3.7 being End-of-Life, update the Python version used for all the services Docker images

Addresses #352 

TODO:

- [x] Figure out if the /opt/v1_0_1 setup in Dockerfile is required or not.
- [ ] Look into why AST parsing linenumber counting changed from python 3.7 -> 3.8+
- [x] Update Github actions to target Python 3.11

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7]
+        python-version: [3.11]
 
     steps:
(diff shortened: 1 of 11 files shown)


saikonen:
still no insights on why the AST function def line numbers change, but this is not a blocker for merging as the only use on the service side is for displaying a tooltip on the Metaflow-UI DAG page. Can be looked into later in a separate PR

PR #370, opened 2023-06-27
fix: S3 access for MinIO resources

saikonen:
Update Metaflow client library version to avoid side-effects during service release build

### DIFF
--- a/services/ui_backend_service/requirements.txt
+++ b/services/ui_backend_service/requirements.txt
@@ -6,7 +6,7 @@ psycopg2
 aiopg
 pygit2==1.6.1
 aiohttp_cors==0.7.0
-metaflow>=2.7.7
+metaflow==2.9.7
 click==8.0.3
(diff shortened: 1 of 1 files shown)


services/ui_backend_service/requirements.txt
```diff
@@ -6,7 +6,7 @@ psycopg2
 aiopg
 pygit2==1.6.1
 aiohttp_cors==0.7.0
-metaflow>=2.7.7
+metaflow==2.9.7
```
savingoyal, line 9: any reason for an exact pin?

saikonen: The MinIO access issue was caused by a bug in the Metaflow client v2.9.5, which was grabbed as part of the services 2.4.0 release due to the loose pin. Having an exact pinned version would avoid this, and make us be deliberate about updating the client dependency only when necessary

PR #373, opened 2023-07-17
Configurable SSL Connection

RikishK:
Adding optional SSL connection.
Currently the database that you can setup for metadata service to use cannot have ssl enabled on it as metadata service cannot make the appropriate ssl connection, this PR offers some new environment variables to allow metadata service to make the appropriate ssl connection to the database. 

If users do not wish for SSL connection, the normal behaviour is still functional.

New ENV Variables:
MF_METADATA_DB_SSL_MODE : If this is set metadata service will try to establish a ssl connection to the db using the following env variables. This env variable can only be set to the following: ['allow', 'prefer', 'require', 'verify-ca', 'verify-full']

MF_METADATA_DB_SSL_CERT_PATH : Path to cert file
MF_METADATA_DB_SSL_KEY_PATH : Path to key file
MF_METADATA_DB_SSL_ROOT_CERT_PATH : Path to CA


services/utils/__init__.py
```diff
@@ -248,18 +256,46 @@ def _is_valid_dsn(dsn):
...
+            if self._ssl_cert_path is not None:
+                ssl_query = f'{ssl_query}&sslcert={self._ssl_cert_path}'
+            if self._ssl_key_path is not None:
+                ssl_query = f'{ssl_query}&sslkey={self._ssl_key_path}'
+            if self._ssl_root_cert_path is not None:
+                ssl_query = f'{ssl_query}&sslrootcert={self._ssl_root_cert_path}'
+        else:
+            ssl_query = f'sslmode=disable'
+
+        print(f'{base_url}?{ssl_query}')
```
RikishK, line 272:
```suggestion
```

savingoyal:
@RikishK can you provide more context on this change?

RikishK:
@savingoyal I added some extra stuff to the description. Please let me know if there is anything else to add.

.github/workflows/test-ssl.yml
```diff
@@ -0,0 +1,118 @@
+name: Test-SSL
```
saikonen, line 1: can we cover the SSL testing in the existing test suite run instead if it is even necessary to introduce tests for this?

RikishK: Hey, the normal testing suite doesn't have ssl for the postgres. This test uses a different postgres image: ghcr.io/infrastructure-as-code/postgres:11 which allows for ssl.

Since the ssl config is optional and does not have to be enabled, I think it should be kept as seperate test. Although I added the test to keep in line with the pattern, unsure if its entirely necessary to have workflow test for ssl connection.

saikonen: After thinking about this for a bit, I would refrain from testing the actual ssl connectivity in the test suite altogether as this is more of a database level implementation detail, rather than something the metaflow-service has to care about.

For testing it should be enough to cover that a correct dsn is created from a DBConfiguration, when SSL related environment variables are present. It is up to the client library and DB then how they choose to handle the connection string.

services/utils/tests/__init__.py
```diff
@@ -1,14 +1,32 @@
...
+        if ssl_cert_path is not None:
+            ssl_query = f'{ssl_query} sslcert={ssl_cert_path}'
+        if ssl_key_path is not None:
+            ssl_query = f'{ssl_query} sslkey={ssl_key_path}'
+        if ssl_root_cert_path is not None:
+            ssl_query = f'{ssl_query} sslrootcert={ssl_root_cert_path}'
+    else:
+        ssl_query = f'sslmode=disable'
+
+    if db_conf.dsn != f"{expected_dsn} {ssl_query}":
```
saikonen, line 29: can be reverted if we agree on the testing scope needed for the ssl feature. This is to establish a connection to a test DB during test suite runs, and the test DB shouldn't be utilizing ssl unnecessarily.

.github/workflows/test-ssl.yml
saikonen: After thinking about this for a bit, I would refrain from testing the actual ssl connectivity in the test suite altogether as this is more of a database level implementation detail, rather than something the metaflow-service has to care about.

For testing it should be enough to cover that a correct dsn is created from a DBConfiguration, when SSL related environment variables are present. It is up to the client library and DB then how they choose to handle the connection string.

RikishK: Sounds good

services/utils/tests/unit_tests/utils_test.py
```diff
@@ -122,11 +122,10 @@ def test_db_conf_env_dsn():
         # Should use default dsn with invalid dsn in environment
         assert DBConfiguration().dsn == 'dbname=postgres user=postgres host=localhost port=5432 password=postgres'
 
-    with set_env({'MF_METADATA_DB_DSN': 'dbname=testgres user=test_user host=test_host port=1234 password=test_pwd'}):
+    with set_env({'MF_METADATA_DB_DSN': 'dbname=testgres user=test_user host=test_host port=1234 password=test_pwd sslmode=disable'}):
```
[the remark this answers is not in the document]
saikonen: Can you add one more test case which covers the ssl related environment variables and verifies that a correct DSN is formed.

services/utils/__init__.py
```diff
@@ -247,19 +255,44 @@ def _is_valid_dsn(dsn):
...
-                host=self._host,
-                port=self._port,
-                password=self._password
-            )
+            ssl_mode = self._ssl_mode
+            sslcert = self._ssl_cert_path
+            sslkey = self._ssl_key_path
+            sslrootcert = self._ssl_root_cert_path
+            if (ssl_mode not in ['allow', 'prefer', 'require', 'verify-ca', 'verify-full']):
+                ssl_mode = 'disable'
```
saikonen, line 280: Tests are currently not passing due to some assertions for the integration tests being made, and this change is appending `sslmode=disable` to the DSN. A quick fix would be to instead set `ssl_mode=None` here to keep the existing functionality intact.

I'm also wondering whether explicitly passing `sslmode=disable` will be a subtle change in behavior, as I checked and the default value for the sslmode in postgres seems to be `prefer` instead of `disable`.

RikishK:
@saikonen Could you run the tests again please, sorry we cant run public actions on our end.

saikonen:
had to introduce a small fix to `run_goose.py` after our test deploy failed with missing ssl env vars, otherwise this is done, thank you for the contribution :)

PR #379, opened 2023-08-14
Upgrade Github actions used in `dockerimage` action

tfurmston:
This `dockerimage` action in this repository is using old versions of `checkout` & `build-push-action`. 

This is causing some warnings during the actions. (See [here](https://github.com/Netflix/metaflow-service/actions/runs/5685654496) for more details.) This includes a warning `The following actions uses node12 which is deprecated and will be forced to run on node16`.

This PR updates the versions of `checkout` & `build-push-action` used in the `dockerimage` action. 

Due to the age of the old versions, to obtain like-for-like functionality the upgrade requires the additional use of `metadata-action`, `setup-buildx-action` & `login-action` actions.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -10,11 +10,11 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       -
         name: Docker meta
         id: meta
-        uses: docker/metadata-action@v4
+        uses: docker/metadata-action@818d4b7b91585d195f67373fd9cb0332e31a7175 # v4.6.0
         with:
           images: |
             netflixoss/metaflow_metadata_service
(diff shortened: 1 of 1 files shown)


savingoyal (approved):

.github/workflows/dockerimage.yml
```diff
@@ -2,22 +2,40 @@ name: Docker Image CI
...
   build:
-
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2.7.0
-      - uses: docker/build-push-action@3e7a4f6646880c6f63758d73ac32392d323eaf8f # v1.1.2
+      -
+        name: Checkout
+        uses: actions/checkout@v3
```
saikonen, line 13: could you also pin the actions to hashes instead of versions like they used to be, as this further mitigates the risks regarding the release workflow.

.github/workflows/dockerimage.yml
```diff
@@ -2,22 +2,40 @@ name: Docker Image CI
...
+        with:
+          images: |
+            netflixoss/metaflow_metadata_service
+          tags: |
+            type=semver,pattern={{raw}}
+            type=sha
+            type=raw,value=latest
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
```
saikonen, line 27: is buildx a required step here since we only target the default `linux/amd64` platform here?

saikonen (changes requested):

.github/workflows/dockerimage.yml
tfurmston: It's a good point. I forgot to pin the versions to patch versions. Very happy to do that. 

One follow up question though. I believe it is possible to use the semantic versions directly, e.g., `uses: actions/checkout@v3.2.1`,  instead of using hashes. Do you have a preference on which to use?

.github/workflows/dockerimage.yml
tfurmston: Yeah, I was also not 100% sure whether to include this step. 

It's not strictly required, but the [documentation](https://github.com/docker/build-push-action#usage) of the `build-push-action` recommends its usage. 

Happy to remove it. Would you prefer it removed or kept?

.github/workflows/dockerimage.yml
saikonen: hashes are preferred to semantic versions, as hashes are static. You can use https://app.stepsecurity.io/secureworkflow/ to easily pin the hashes based on the versions supplied, instead of having to dig for them manually :)

.github/workflows/dockerimage.yml
tfurmston: cool. That's done now.

.github/workflows/dockerimage.yml
tfurmston: Cool. Makes sense. 

That's done now.

PR #381, opened 2023-08-17
[ui] config root configurable by env vars

valayDave:


services/ui_backend_service/api/utils.py
```diff
@@ -38,7 +38,7 @@ def get_json_config(variable_name: str):
     env_name = variable_name.upper()
 
     filepath = os.path.join(JSON_CONFIG_ROOT, f"config.{variable_name.lower()}.json")
-
+    logger.info("Looking for JSON config in env: {} or file: {}".format(env_name, filepath))
```
rohanrebello, line 41: An f-string would have been nice here...

rohanrebello: https://github.com/Netflix/metaflow-service/pull/382, more so that I get some practice with the OSS pull request and release model

PR #388, opened 2023-09-05
fix PATH_PREFIX handling in metadata service so it doesn't interfere with mfgui routes

oavdeev:
follow up fix for https://github.com/Netflix/metaflow-service/pull/384 , without this it can break MFGUI


services/metadata_service/server.py
```diff
@@ -34,16 +34,16 @@ def app(loop=None, db_conf: DBConfiguration = None, middlewares=None):
     ArtificatsApi(app)
     AuthApi(app)
 
-    if len(PATH_PREFIX) > 0:
-        _app.add_subapp(PATH_PREFIX, app)
+    if path_prefix:
```
saikonen, line 37: would prefer explicitly checking for the default `if path_prefix is not None`

services/metadata_service/server.py
```diff
@@ -18,12 +18,12 @@
 PATH_PREFIX = os.environ.get("PATH_PREFIX", "")
```
saikonen, line 18: change default to `None` instead of empty string to match the default value for prefix in `app()` ?

saikonen:
for context why the existing changes would break MFGUI when using `PATH_PREFIX`: The ui_backend imports metadata service`app()` and hosts it as a subapp. Therefore the setting of the prefix config needs to be kept outside of the app function in the metadata service.

saikonen (approved):

services/metadata_service/server.py
oavdeev: Hmm I didn't want to treat "None" vs "" differently in any way here. Let me just change the function argument default to `""` for symmetry

saikonen (approved):

PR #392, opened 2023-09-20
fix: tone down run inactive cutoff default

saikonen:
Changes the Run inactive cutoff time default to 6 minutes from 1 day. After the latest query changes this is used to determine whether a run heartbeat is expired or not.
This will give runs that have tasks stuck in a scheduler the 6 minutes grace period for any of them to launch, after which the run level heartbeat will expire and mark the run as failed. If a task starts after this, the run would flip back to `running` status, otherwise remaining `failed` (correctly)

Having to wait a whole day for runs to be marked as failed by default is such a huge departure from the way the UI used to display failed runs that this change is proposed as a possible middle-ground instead of reverting the query changes from #338 .


saikonen:
another option that has been thought of is to make the heavy query logic prior to #338 conditional behind an env var. This will complicate the code significantly though, but would allow both solutions to coexist and be configurable based on deployment needs.

PR #394, opened 2023-10-20
Speed up log lookup in UI backend

martinbattentive:
We've run into issues with the STDOUT/STDERR log view feature in the
Metaflow UI.  The current code path does an expensive join query to
determine the latest attempt, which ends up timing out.  However, the
relevant request is already supplying the attempt_id provided by the UI
from an earlier query, so there is no need to perform this verification.

Additional unrelated change: one of the existing tests was failing due
to the async duration not falling within the expected tolerance.  This
tolerance appeared to be specified incorrectly, using '1000' as a
relative tolerance.  I assume this was meant to be absolute tolerance,
as in 'this request should take (0 +- 1000 ms) to complete."  In my
development environment I needed 5000 to guarantee success with the
async nature of the tests.

### DIFF
--- a/services/ui_backend_service/api/log.py
+++ b/services/ui_backend_service/api/log.py
@@ -1,7 +1,6 @@
 
-from typing import Optional, Tuple
 
-from services.data.db_utils import DBResponse, translate_run_key, translate_task_key, DBPagination, DBResponse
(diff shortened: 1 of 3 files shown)


martinbattentive:
Was related to the log CacheAsyncClient getting into a bad state and becoming unresponsive.  Restart of ui_backend service resolved the issue.

PR #402, opened 2023-11-26
[realtime-cards] backend API Methods

valayDave:


services/ui_backend_service/api/card.py
```diff
@@ -133,20 +141,113 @@ async def get_card_content_by_hash(self, request):
...
+        await res.wait()  # wait until results are ready
+    data = res.get()
+    if data and pathspec in data:
+        success, value, detail, trace = unpack_processed_value(data[pathspec])
+        if success:
+            return value
+        else:
+            raise CardException(detail, value, trace)
+
+    return None
```
valayDave, line 245: Meat of the changes

services/ui_backend_service/api/card.py
```diff
@@ -133,20 +141,113 @@ async def get_card_content_by_hash(self, request):
...
+            return web.Response(
+                content_type="text/html", status=404, body="Task not found."
+            )
+
+        data = await get_card_data_for_task(self.cache, _hash, task)
+
+        if data is None:
+            return web_response(404, {"error": "Card data not found for task"})
+        else:
+            return web_response(200, data)
```
valayDave, line 202: Meat of the changes

services/ui_backend_service/data/cache/get_card_data_action.py
valayDave, line 1: Meat of the changes

services/ui_backend_service/data/cache/store.py
```diff
@@ -145,7 +146,7 @@ async def restart_requested(self):
 
     async def start_cache(self):
         "Initialize the CacheAsyncClient for artifact caching"
-        actions = [GetData, SearchArtifacts, GetTask, GetArtifacts, GetParameters, GetCards]
+        actions = [GetData, SearchArtifacts, GetTask, GetArtifacts, GetParameters, GetCards, GetCardData]
```
valayDave, line 149: Meat of the changes

services/ui_backend_service/data/cache/get_card_data_action.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Stream error example:
+            stream_output(error_event_msg(str(ex), "s3-not-found", get_traceback_str()))
+        """
+
+        def _card_item(card):
+            card_load_policy = os.environ.get("MF_CARD_LOAD_POLICY", "full")
+            card_data = None
+            if card_load_policy == "full":
+                card_data = card.get_data()
```
saikonen, line 61: `.get_data` being added in https://github.com/Netflix/metaflow/pull/1550 so this service PR should only go in once the client changes are released. We also need to pin the Metaflow client version to a new minimum for the service release

services/ui_backend_service/api/card.py
```diff
@@ -133,20 +141,113 @@ async def get_card_content_by_hash(self, request):
...
+        ---
+        description: Get the data of a card created for a task. Contains any additional updates needed by the card.
+        tags:
+        - Card
+        parameters:
+          - $ref: '#/definitions/Params/Path/flow_id'
+          - $ref: '#/definitions/Params/Path/run_number'
+          - $ref: '#/definitions/Params/Path/step_name'
+          - $ref: '#/definitions/Params/Path/task_id'
+          - $ref: '#/definitions/Params/Custom/invalidate'
```
saikonen, line 176: doesn't actually do anything with the invalidate query param

saikonen (commented):
changes seem straightforward enough to go in once the client side is finalized. Are there going to be additional changes to the UI to support realtime updates for the cards content? Is the plan to subscribe to metadata events for a task, or polling for the card data periodically?

services/ui_backend_service/api/card.py
```diff
@@ -133,20 +141,113 @@ async def get_card_content_by_hash(self, request):
...
+    pathspec = "{flow_id}/{run_id}/{step_name}/{task_id}".format(
+        flow_id=task.get("flow_id"),
+        run_id=task.get("run_id") or task.get("run_number"),
+        step_name=task.get("step_name"),
+        task_id=task.get("task_name") or task.get("task_id"),
+    )
+    # Invalidate cache until the task completes because the card data may keep getting updated
+    # Ideally status should NEVER be NOT_SET
+    task_status = task.get("status", "NOT_SET")
+    invalidate_cache = task_status in ["pending", "running", "NOT_SET"]
```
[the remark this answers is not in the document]
valayDave: Will Fix!

services/ui_backend_service/data/cache/get_card_data_action.py
valayDave: Will add a dependency pin!

services/ui_backend_service/api/card.py
valayDave: Will Fix!. Bad copy pasta

valayDave:
@saikonen : thank you for the review. 
> Are there going to be additional changes to the UI to support realtime updates for the cards content? 

Yes, @obgibson is helping make the changes on the UI side. 

>Is the plan to subscribe to metadata events for a task, or polling for the card data periodically?

The UI logic will be polling based on the lifecycle of a single card:

Once a  card is loaded in Iframe: Check if the iframe window contains `metaflow_card_update` function and `METAFLOW_RELOAD_TOKEN` property; If not: Do nothing. `METAFLOW_RELOAD_TOKEN` is a string which will be set at the window level in the iframe. If the iframe contains both these properties, then at a periodic interval :

1. call the card-data API: (One added in this PR)
2. The API response will contain a reload_token
3. if `response.payload.reload_token==iframe.window.METAFLOW_RELOAD_TOKEN` : call the `metaflow_card_update` function in the iframe with the data : `iframe.window.metaflow_card_update(response.payload.data)`
4. else: Re-fetch the card from the server by calling the card fetch API route. Load the card in the Iframe and check if the iframe window contains `metaflow_card_update` function and `METAFLOW_RELOAD_TOKEN` property. If it does, then do nothing (Let the polling continue); If it doesn't then remove the `setInterval` which runs the logic from step 1 at a periodic interval.

valayDave:
@saikonen: i will have a look at the test failures in the next 24 hours too; I will ping you once on the change of the test since we have some logic change in the `/cards` API too given that we may have to invalidate cache more often than not.

PR #408, opened 2023-12-28
fix: change ssl_mode to sslmode

rswigginton:
Fixes sslmode query parameter typo in goose migrations script.


joshcx:
Bumping this. Facing the following issue due to the typo:

`psycopg2.errors.UndefinedTable: relation "public.flows_v3" does not exist`

saikonen (approved):

PR #410, opened 2024-01-11
[realtime-cards] fix bug in mf version

valayDave:


valayDave:
Tests failing coz Metaflow is still not published 🤷 https://github.com/Netflix/metaflow/actions/runs/7493205983

rohanrebello (approved):
Verified that https://pypi.org/project/metaflow/2.10.9/ is the latest version

PR #414, opened 2024-02-07
fix: make search api changes backwards compatible

saikonen:
small change to make search api backward compatible with older UI versions


darinyu (approved):
Thanks for adding this!

PR #416, opened 2024-02-16
[cache-memory-leak] Fix Memory leak in cache server

valayDave:
## Key changes
- Recreate the multiprocess pool at a regular cadence to avoid memory leaks
- Since the pool was never removed it resulted in unbounded growth of memory.
- Added a log size constraint to the `tail` based Log cache setting.

### DIFF
--- a/services/ui_backend_service/data/cache/client/cache_server.py
+++ b/services/ui_backend_service/data/cache/client/cache_server.py
@@ -9,9 +9,12 @@ import multiprocessing
 from datetime import datetime
 from collections import deque
 from itertools import chain
+import time
 
 from .cache_worker import execute_action
 from .cache_async_client import OP_WORKER_CREATE, OP_WORKER_TERMINATE
+
+import sys
 
 import click
 
@@ -30,6 +33,10 @@ def send_message(op: str, data: dict):
         'op': op,
         **data
     }), flush=True)
+
+
+CACHE_PROCESS_POOL_REFRESH_DURATION = int(os.environ.get("CACHE_PROCESS_POOL_REFRESH_DURATION", 20 * 60))
+CACHE_PROCESS_POOL_FORCE_REFRESH_DURATION = int(os.environ.get("CACHE_PROCESS_POOL_FORCE_REFRESH_DURATION", 2 * 60))
 
 
 class CacheServerException(Exception):
@@ -101,6 +108,7 @@ class Worker(object):
         self.pool = pool
         self.callback = callback
         self.error_callback = error_callback
+        self.created_on = time.time()
 
         try:
             self.tempdir = self.filestore.open_tempdir(
@@ -198,6 +206,7 @@ class Scheduler(object):
             initializer=self.init_process,
             maxtasksperchild=512,  # Recycle each worker once 512 tasks have been completed
         )
+        self._pool_started_on = time.time()
 
     def init_process(self):
         echo("Init process %s pid: %s" % (multiprocessing.current_process().name, os.getpid()))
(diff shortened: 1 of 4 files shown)


services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -257,16 +265,64 @@ def queued_request(queue):
...
+            "number of workers: %d, number of pending requests: %d; Pool Refresh in : %d" % (
+                len(self.workers), len(self.pending_requests), time_to_pool_refresh)
+        )
+
+    def cleanup_if_necessary(self):
+        time_to_pool_refresh = CACHE_PROCESS_POOL_REFRESH_TIME - (time.time() - self._pool_started_on)
+        if time_to_pool_refresh > 0:
+            return
+        # if workers are still running 30 seconds after the pool refresh timeout, then cleanup
+        no_workers_are_running = len(self.workers) == 0 and len(self.pending_requests) == 1
```
romain-intel, line 281: does this mean an idle pool has one request?

services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -257,16 +265,64 @@ def queued_request(queue):
...
+        pool_needs_refresh = time_to_pool_refresh <= 0
+        pool_force_refresh = time_to_pool_refresh < -60
```
romain-intel, line 283: Do we want to make that configurable?

services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -257,16 +265,64 @@ def queued_request(queue):
...
+        if pool_force_refresh:
+            echo("Refreshing the pool as no workers are running and no pending requests are there.")
+            self.cleanup_workers()
+        elif no_workers_are_running and pool_needs_refresh:
+            echo("Refreshing the pool as no workers are running and no pending requests are there.")
```
romain-intel, line 288: is there no race here? Can a worker not "get busy" in the meantime?

services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -257,16 +265,64 @@ def queued_request(queue):
...
+        _counter = time.time()
 
         while True:
             self.process_incoming_request()
             new_worker_from_request()
+            if time.time() - _counter > 30:
+                self.verify_stale_workers()
+                _counter = time.time()
+
+            self.cleanup_if_necessary()
```
romain-intel, line 325: we could probably put this in the if statement. Not a huge difference though.

romain-intel (approved):
Minor nits but LGTM.

services/ui_backend_service/data/cache/client/cache_server.py
valayDave: Do you mean the force refresh setting ? Yes, I can make it configurable.

services/ui_backend_service/data/cache/client/cache_server.py
valayDave: Given that this entire execution is running _sync_ code, and this code block comes after the code path that creates new workers, the only current case is that workers are already created. No _new_  worker will be created during the execution of this block. If workers exist, they are already busy workers. 

Generally, if workers are busy at the time of pool-refresh, I didn't want to kill them immediately. If the server is under a lot of load, then an API request on the server may be waiting for some object from the server whose worker may get killed, and will result in the API response being an ugly http-500x.  

I have not exactly explained in this PR description how the memory leak came to be so let me more light: 
1. When API requests come to the server, the UI Server request the cache-server (running in the background) to extract out the cached value. This request is given a stream-id. The API-server is waiting on the response of the stream-id from the cache-server.
2. The cache server spawns a new worker for this stream-id and worker does some computing and returns the object in the stream
3. When many requests hit the server (especially when a run has a lot of logs), the cache server will spawn multiple individual workers, The API server will be waiting on all of these streams. If the objects(ex. logs) are large, then it will end up loading the whole object in memory and give the response. After the response is passed, the callback handles the removal of the worker and committing information to the cache-store. Even if the worker has been liked, the pool is still holding the reference to it (since we never close/join the pool)

services/ui_backend_service/data/cache/client/cache_server.py
valayDave: yes, that is what I observed.

services/ui_backend_service/data/cache/client/cache_server.py
```diff
@@ -257,16 +265,64 @@ def queued_request(queue):
...
             time.sleep(0.1)
```
valayDave, line 326: In Theory, this sleep could be smaller.

services/ui_backend_service/data/cache/get_log_file_action.py
```diff
@@ -206,16 +224,21 @@ def get_log_content(self, task: Task, logtype: str):
...
 
     def get_log_content(self, task: Task, logtype: str):
-
+        log_size = get_log_size(task, logtype)
+        if log_size > self._max_log_size:
+            return [(
+                None, log_size_exceeded_blurb(task, logtype, self._max_log_size)
+            ), ]
         # Note this is inefficient - we will load a 1GB log even if we only want last 100 bytes.
         # Doing this efficiently is a step change in complexity and effort - we can do it when justified in future.
```
rohanrebello, line 243: I guess we've arrived at that future point where it's justified :)

PR #420, opened 2024-04-15
[card-cache] optimizations to avoid deadlocks

valayDave:
[card-cache] optimizations to avoid deadlocks
- deadlocks happened when cleanups coincided with heavy load 
- tests simulating a cleanup (disk + shared-objects) and heavy load together were successfully able to reproduce the deadlock situation
- Locking at the `context` level: 
    - Each time we want to clean up we lock to create a new context
    - All directories/processes are written within that new context.
    - Switching this context which ensure that all new processes get created differently and the cleanup process can safely remove everything. 
    - Context also sets the read/write directory for the cache object used in the API endpoint.
    - All locking on the API side now always time-bound. The code time's out if it can't acquire a lock. 
    - The method ensure that all operations won't in-definately hold the lock.
- Changed defaults for minimum amount to time to wait for cards in the cache process to 20 seconds (helps make things snappier)
- Added `timings` dict in card cache to optimize loading cycles (Ensured that it is set based on a per-card basis)

### DIFF
--- a/services/ui_backend_service/api/card.py
+++ b/services/ui_backend_service/api/card.py
@@ -153,11 +153,12 @@ class CardsApi(object):
             task,
             hash,
         )
+        html_reload_script = "<script>window.needsReload = true;</script><body></body>"
         if cards is None:
             return web.Response(
                 content_type="text/html",
                 status=404,
-                body="Card not found for task. Possibly still being processed. Please refresh page to check again.",
+                body=html_reload_script,
             )
 
         if cards and hash in cards:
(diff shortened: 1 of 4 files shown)


services/ui_backend_service/api/card.py
```diff
@@ -153,11 +153,12 @@ async def get_card_content_by_hash(self, request):
             task,
             hash,
         )
+        html_reload_script = "<script>window.needsReload = true;</script><body></body>"
         if cards is None:
             return web.Response(
                 content_type="text/html",
                 status=404,
-                body="Card not found for task. Possibly still being processed. Please refresh page to check again.",
+                body=html_reload_script,
```
valayDave, line 161: Change to UI coming soon!.

valayDave:
#442 supersedes this PR

PR #423, opened 2024-05-23
feature: support streaming large logs

saikonen:
adds support for streaming log file downloads to be able to support handling large log files.
Depends on https://github.com/Netflix/metaflow/pull/1856

Development environment changes:
- mounts .aws folder in development docker-compose in order to provide S3 access for the local UI backend

TODO:

- [ ] cleanup dependency changes due to temporarily installing the Metaflow client from a git branch

PR #424, opened 2024-06-05
feature: massive logs streaming

saikonen:
Alternative approach to streaming massive logs. Does not require any changes to Metaflow client side.

Intended to supersede #423 

TODO:

- [x] support legacy logs
- [x] reliable cache/GC of the raw log downloads on disk so they can be reused across cache action calls.

### DIFF
--- a/services/ui_backend_service/data/cache/get_log_file_action.py
+++ b/services/ui_backend_service/data/cache/get_log_file_action.py
@@ -1,9 +1,14 @@
 import hashlib
 import json
 
-from typing import Dict, List, Optional, Tuple
+import shutil
+from typing import Callable, Dict, List, Optional, Tuple
 from .client import CacheAction
 from .utils import streamed_errors
+from metaflow.client.filecache import FileCache
+from metaflow.mflog import LOG_SOURCES
+from metaflow.mflog.mflog import MFLogline, parse, MISSING_TIMESTAMP_STR, MISSING_TIMESTAMP
+from metaflow.util import to_unicode
 import os
 
 # New imports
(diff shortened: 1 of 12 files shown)


services/ui_backend_service/data/cache/get_log_file_action.py
```diff
@@ -192,6 +202,147 @@ def log_size_exceeded_blurb(task: Task, logtype: str, max_size: int):
...
+        # Load the log files to disk
+        with ds._storage_impl.load_bytes(to_load) as load_results:
+            for key, path, meta in load_results:
+                name = ds._storage_impl.basename(key)
+                if path is None:
+                    # no log file existed in the location. Usually not an error.
+                    log_paths[name] = None
+                else:
+                    shutil.move(path, log_paths[name])
+    return [val for val in log_paths.values() if val is not None]
```
romain-intel, line 281: I wonder if this whole method could just be simplified by using the task datastore's load_logs (well through the filecache if you want to although not 100% sure) with the addition of something like "write_to_disk_only". In other words, you seem to be doing a lot of the thing from there with the only difference being that you don't open the logs but just keep them on disk. I feel we could make that small change there (without modifying any of the storage layers but just keeping it at the task datastore level) and this should simplify this code a lot. Basically, your line 280 (or somethin akin to it) would go around line 984 of task_datastore.py (protected by a flag or something).

romain-intel (commented):
A few comments -- i'll follow up offline

services/ui_backend_service/data/cache/get_log_file_action.py
saikonen: yep, as a follow up will look into hoisting all the logic required by fetch_logs into the Metaflow client side, for something akin to `dump_logs(stream, destination_path)`

PR #429, opened 2024-06-17
Change websocket cleanup performance from O(n^2) to O(n).

marshallbrekka:
Move subscription tracking to a dict keyed by the underlying websocket object.

Changes the cleanup from two nested for loops (the 2nd being the unsubscribe_from method), to a single loop with (the common case) deleting a single key from a dict.

We found that with a sufficient number of disconnected clients, and long gaps between database activity, the cleanup could end up taking multiple hours. This would cause the python web service to become unresponsive.


savingoyal (approved):

savingoyal:
lgtm! we can ship this once @saikonen approves

marshallbrekka:
@savingoyal @saikonen is there any update here, anything I can do to help get this upstreamed? thanks!

services/ui_backend_service/api/ws.py
```diff
@@ -108,18 +108,28 @@ async def event_handler(self, operation: str, resources: List[str], data: Dict,
...
                 except Exception:
                     self.logger.exception("Broadcasting to subscription failed")
 
+    def subscriptions(self):
+        # Grab all of the keys upfront and use that to iterate so that callers can
+        # safely modify the subscriptions dict while we are iterating through it.
+        # This is primarily useful when calling `unsubscribe_from` during the cleanup
+        # loop in the event handler.
+        for k in list(self._subscriptions.keys()):
+            for sub in self._subscriptions[k]:
```
saikonen, line 130: note to self: no KeyError possible due to self._subscriptions being a `defaultdict(list)`

services/ui_backend_service/api/ws.py
```diff
@@ -154,22 +164,25 @@ async def subscribe_to(self, ws, uuid: str, resource: str, since: int):
                 )
 
     async def unsubscribe_from(self, ws, uuid: str = None):
+        if ws not in self._subscriptions:
+            return
         if uuid:
-            self.subscriptions = list(
-                filter(lambda s: uuid != s.uuid or ws != s.ws, self.subscriptions))
+            self._subscriptions[ws] = list(
+                filter(lambda s: uuid != s.uuid or ws != s.ws, self._subscriptions[ws]))
```
saikonen, line 171: is the `ws != s.ws` check necessary anymore as the subscription is keyed per websocket?

saikonen (commented):
overall changes look good. had one minor note on the filter part.

linter step is failing due to unrelated reasons, this is fixed in `master` now if you want to rebase

saikonen (approved):

PR #431, opened 2024-08-20
[bug fix] log-streaming

valayDave:


saikonen:
closing in favor of #435 which includes unit tests for the feature. fix contents should be the same

PR #432, opened 2024-08-22
feature: version table triggers

saikonen:
this adds versioning to table triggers created by the ui_service.

triggers will now have a timestamp as part of their name, and as part of the trigger setup process when launching the ui_backend it will attempt to clean up table triggers and trigger functions that do not have the correct timestamp (or lack one completely) but otherwise match our trigger naming schema.

Custom triggers not set up by the ui_backend will remain untouched by the changes

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -38,6 +38,8 @@ DB_SCHEMA_NAME = os.environ.get("DB_SCHEMA_NAME", "public")
 
 operator_match = re.compile('([^:]*):([=><]+)$')
 
+# use a ddmmyyy timestamp as the version for triggers
+TRIGGER_VERSION = "18012024"
 
 class _AsyncPostgresDB(object):
     connection = None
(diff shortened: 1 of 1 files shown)


services/data/postgres_async_db.py
```diff
@@ -428,12 +442,52 @@ async def create_trigger_if_missing(db: _AsyncPostgresDB, table_name, trigger_na
...
+                    (res[0], res[1]) for res in results
+                    if res[0].startswith(TRIGGER_NAME_PREFIX) and TRIGGER_VERSION not in res[0]
+                ]
+                if triggers_to_cleanup:
+                    logging.getLogger("TriggerSetup").info("Cleaning up old triggers: %s" % triggers_to_cleanup)
+                    commands = []
+                    for trigger_name, schema in triggers_to_cleanup:
+                        commands += [
+                            (f"DROP TRIGGER IF EXISTS {trigger_name} ON {table_name}"),
+                            (f"DROP FUNCTION IF EXISTS {schema}.{trigger_name}")
```
saikonen, line 469: @romain-intel This is the only major difference to the setup in the previous PR

as suggested, I now check for the known prefix for ui-triggers, as well as look for the expected version before deciding if a trigger needs to be cleaned up or not.

I also added dropping the actual trigger function, as previously I was only uncoupling the function from the table.

saikonen: whether the checks are computationally expensive or not is not relevant here, as the code runs only during server startup when triggers are setup.

PR #440, opened 2024-09-09
Fix `AsyncPostgresDB` initialization retry logic

### LINKED ISSUE #439: `AsyncPostgresDB` does not raise exception when initialization fails
`AsyncPostgresDB` initializes a PostgreSQL connection pool. It has retry logic to handle initialization failures; however, there is a bug in that logic that causes the exception to not be thrown after there are no more attempts.

fumoboy007:
Fixes #439.

### DIFF
--- a/services/migration_service/data/postgres_async_db.py
+++ b/services/migration_service/data/postgres_async_db.py
@@ -35,15 +35,18 @@ class AsyncPostgresDB(object):
         AsyncPostgresDB.__instance = self
 
     async def _init(self, db_conf: DBConfiguration):
-        # todo make poolsize min and max configurable as well as timeout
-        # todo add retry and better error message
-        retries = 3
-        for i in range(retries):
+        # todo make poolsize min and max configurable
+        max_attempts = 3
+        while max_attempts > 0:
             try:
                 self.pool = await aiopg.create_pool(db_conf.get_dsn(), timeout=db_conf.timeout)
+                break
             except Exception as e:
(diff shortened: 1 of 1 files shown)


services/migration_service/data/postgres_async_db.py
```diff
@@ -35,15 +35,18 @@ def __init__(self):
...
-        # todo make poolsize min and max configurable as well as timeout
-        # todo add retry and better error message
-        retries = 3
-        for i in range(retries):
+        # todo make poolsize min and max configurable
+        max_attempts = 3
+        while max_attempts > 0:
             try:
                 self.pool = await aiopg.create_pool(db_conf.get_dsn(), timeout=db_conf.timeout)
+                break
```
saikonen, line 43: would it not have been sufficient to just add the break to the original code?

services/migration_service/data/postgres_async_db.py
```diff
@@ -35,15 +35,18 @@ def __init__(self):
...
             except Exception as e:
                 print("printing connection exception: " + str(e))
-                if retries - i < 1:
```
saikonen, line 46: I see this is never actually hit as range excludes the end, but can be fixed with a slight tweak f.ex. 

```
if retries-(i+1) < 1:
```

saikonen (commented):
Thank you for the bug report and fixes.

We can go forward with the current changes as well, but at a quick glance it seemed that this could've been fixed with a smaller diff as well. Could you make the changes I noted?

Issue #337, opened 2022-10-31
IAM Authentication for RDS?


gucciwang:
Hi Metaflow! 

Is iam_authentication for RDS on current roadmap for Metaflow? 

Thanks in advance!

PR #458, opened 2025-06-10
fix: percent-encode forward slash in database connection string

wedaly:
The DB connection string used `urllib.parse.quote` to percent-encode the username and password. However, `quote` has a default kwarg `safe="/"` that does NOT percent-encode forward slashes. If the username or password includes a forward slash, then the resulting connection string may be rejected by the postgres client.

For example, when the password contains an unescaped forward slash, psycopg incorrectly parses the password field as a port.

```
>>> import psycopg2
>>> conn = "postgresql://user:abc/123@localhost:5432/db"
>>> psycopg2.connect(conn)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    psycopg2.connect(conn)
    ~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/willdaly/repos/metaflow-service/.venv/lib/python3.13/site-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: invalid integer value "abc" for connection option "port"
```

Fix it by percent-escaping forward slashes in all DB connection strings.

### DIFF
--- a/run_goose.py
+++ b/run_goose.py
@@ -15,6 +15,7 @@ DB_SCHEMA_NAME = os.environ.get("DB_SCHEMA_NAME", "public")
 # Set safe="" to percent-encode everything.
 quote_all = partial(quote, safe="")
 
+
 def check_if_goose_table_exists(db_connection_string: str):
     conn = psycopg2.connect(db_connection_string)
     cur = conn.cursor()
--- a/run_goose.py
+++ b/run_goose.py
@@ -2,6 +2,7 @@ import os
 import sys
 import time
 import argparse
+from functools import partial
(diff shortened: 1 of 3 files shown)


wedaly:
incidentally, I discovered this while trying to use [AWS RDS IAM](https://github.com/Netflix/metaflow-service/issues/337). AWS puts forward slashes in its tokens.

savingoyal (approved):

savingoyal:
It seems that linter check failed. Can you address that so that we can merge this PR?

PR #461, opened 2026-02-04
feature: non root images

saikonen:
changes Docker images so they run as non-root.

Note that introducing this change moves the location of `run_goose.py`, which is a breaking change: This script is directly referenced in deployment templates for the services, so updating those is a requirement as well.


schack:
It would be really nice to have this one merged

PR #463, opened 2026-02-20
feat(api): cursor-based pagination for GET /flows/{flow_id}/runs (#322)

### LINKED ISSUE #322: `500` error encountered on `flows/<flow_id>/runs` requests due to size of payload
I have a flow running in production on a `2min` schedule for which artifacts / metadata are being stored. I am trying to interact with those artifacts with the [client API](https://docs.metaflow.org/metaflow/client). However, requests to  `/flows/<flow_id>/runs` (via instantiating `Flow("FlowName")`) are now failing with
```
Metadata request: (/flows/<flow_id>/runs) failed (code 500): {"message": "Internal server error"}
```
Request to other endpoints, like `flows/<flow_id>/runs/<run_id>`, are going through just fine. After taking a closer look on API Gateway, I generated the same request through the console there, and get
```
Execution failed due to configuration error: Integration response of reported length 28729085 is larger than allowed maximum of 10485760 bytes.
Tue May 31 22:49:46 UTC 2022 : Method completed with status: 500
```
Basically the response payload is larger than the non-configurable [10MB limit](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html) on API Gateway. 

I can get around this by requesting individual runs directly, but it would be great to still be able to use the `flow` apis to interact with the child runs of the flow via the client. (Perhaps adding the ability to pass filtering params in the requests, like fetching the last `n` runs?). Also curious if this is something that others have run into when deploying flows to production where many runs are produced and stored, and if there are any workarounds or thing I am missing.

GunaPalanivel:
## Description

- Fixes #322

This PR adds cursor-based pagination to `GET /flows/{flow_id}/runs` using `run_number` as the cursor.

Response body shape is unchanged (still a flat array of runs).  
When pagination is enabled, metadata is returned through headers:

- `Link` with `rel="next"` (RFC 5988 style)
- `X-Total-Count`
- `X-Pagination-Limit`

## Backward Compatibility

Backward compatibility is preserved by default:

- If `_limit` is omitted, endpoint keeps legacy behavior and returns all runs without pagination headers.
- If `_limit` is provided, it must be a positive integer and pagination is applied.
- `_limit=0` is now treated as invalid (`400`) to avoid ambiguous semantics.

Also, `_after` is only valid when `_limit` is provided.

## Problem

`/flows/{flow_id}/runs` currently returns all runs in one unbounded response.  
For large flows this can lead to:

- API Gateway payload-size failures
- high latency
- unnecessary memory pressure
- no way to consume results incrementally

### Cursor Choice

`run_number` is monotonic and stable for pagination.

- order is `run_number DESC` (newest first)
- next cursor is the last run_number from the current page
- `Link` is emitted only when current page is full (possible next page)

### Count Behavior and Trade-off

`X-Total-Count` is produced with a separate `COUNT(*)` query scoped by `flow_id`.

- For current query shape, this is acceptable and index-backed.
- For future tag-based JSONB filters, we will need to revisit strategy (for example GIN index and/or different count behavior when tag filters are active).

`X-Total-Count` is best-effort metadata, not a strict transactional guarantee under concurrent writes.

## Validation and Errors

Returns `400` for invalid pagination inputs:

- non-integer `_limit`
- `_limit <= 0`
- non-integer `_after`
- `_after <= 0`
- `_after` provided without `_limit`

## Tests

Integration coverage includes:

- omitted `_limit` -> legacy unbounded behavior, no pagination headers
- `_limit=2` -> multi-page traversal with `Link` next and no duplicates
- ordering remains newest-first
- invalid params return `400` (including `_limit=0`)
- non-existent flow returns `200` with empty array

## Scope

This PR is intentionally scoped to runs endpoint pagination only.  
Runs are the endpoint with unbounded growth and immediate payload-risk; other resources can be handled in follow-ups if needed.

## AI Tool Usage

AI assistance was used during implementation.  
All changes were reviewed manually, validated locally, and adjusted based on reviewer feedback.

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -220,8 +220,13 @@ class AsyncPostgresTable(object):
         Return the total count of rows matching filter_dict.
         Used for X-Total-Count pagination header without fetching all rows.
         Uses reader pool when available (read-only query).
-        Returns 0 on any DB error rather than propagating — callers treat
+        Returns 0 on any DB error rather than propagating -- callers treat
         X-Total-Count as a best-effort hint, not a guarantee.
+
+        Column names in filter_dict are validated against self.keys before
+        interpolation; unknown columns are skipped to prevent SQL injection.
+        JSONB fields (e.g. tags) are intentionally not supported here --
+        tag-aware counting would need a GIN index and a separate strategy.
         """
         if filter_dict is None:
             filter_dict = {}
@@ -655,7 +660,7 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
     async def get_all_runs_paginated(
         self,
         flow_id: str,
-        limit: int = 200,
+        limit: int,
         after_run_number: int = None,
     ) -> Tuple["DBResponse", int]:
         """
@@ -670,7 +675,21 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
         (DBResponse, total_count)
             DBResponse.body is a flat list of run dicts (unchanged schema).
             total_count is the full run count for the flow (for X-Total-Count
-            header) — computed independently of the cursor.
+            header) -- computed independently of the cursor.
+
+        Performance notes
+        -----------------
+        The COUNT query is `SELECT COUNT(*) FROM runs_v3 WHERE flow_id = %s`.
+        PostgreSQL can satisfy this with an index-only scan on the
+        (flow_id, run_number) primary key B-tree, so it is fast for the
+        current query shape.
+
+        If future work adds tag-based filters (e.g. `tags @> '{...}'::jsonb`),
+        that condition cannot use the B-tree index. It would require a GIN
+        index on `tags || system_tags` and the count strategy should be
+        revisited at that point -- for example, skipping X-Total-Count
+        entirely when tag filters are active, or using a separate estimate.
+        That work is tracked as a follow-up and is out of scope here.
         """
         conditions = ["flow_id = %s"]
         values = [flow_id]
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -215,6 +215,48 @@ class AsyncPostgresTable(object):
         )
         return response
 
+    async def count_records(self, filter_dict=None) -> int:
+        """
+        Return the total count of rows matching filter_dict.
+        Used for X-Total-Count pagination header without fetching all rows.
+        Uses reader pool when available (read-only query).
+        Returns 0 on any DB error rather than propagating — callers treat
+        X-Total-Count as a best-effort hint, not a guarantee.
+        """
+        if filter_dict is None:
+            filter_dict = {}
+        conditions = []
+        values = []
+        for col_name, col_val in filter_dict.items():
+            # Validate column name against known keys to prevent SQL injection
+            if col_name not in self.keys:
+                self.db.logger.warning("count_records: unknown column %s", col_name)
+                continue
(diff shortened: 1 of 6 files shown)


GunaPalanivel:
@romain-intel could you trigger CI when you get a chance?

All integration tests pass locally (8 passed, 0 regressions).
Happy to address any review feedback.

services/data/postgres_async_db.py
```diff
@@ -215,6 +215,48 @@ async def get_records(self, filter_dict={}, fetch_single=False,
         )
         return response
 
+    async def count_records(self, filter_dict=None) -> int:
```
saikonen, line 218: any performance concerns with the `COUNT` approach for providing pagination data? there is a plan to introduce filtering based on tags as well, which would require matching values in the JSONB tags column.

services/metadata_service/api/run.py
saikonen, line 1: Is it not necessary to add pagination to anything besides runs?

services/metadata_service/api/run.py
```diff
@@ -73,16 +73,92 @@ async def get_all_runs(self, request):
...
+                after_run_number = int(raw_after)
+                if after_run_number <= 0:
+                    return DBResponse(response_code=400, body="Invalid _after: must be a positive run_number")
+            except (ValueError, TypeError):
+                return DBResponse(response_code=400, body="Invalid _after: must be an integer run_number")
+        else:
+            after_run_number = None
+
+        # --- Legacy opt-out (_limit=0): unbounded query, no headers -------
+        if limit == 0:
```
saikonen: how do we ensure backwards compatibility here? it seems the limit is never `0` if the client does not explicitly set it to the value.

services/data/postgres_async_db.py
GunaPalanivel: In this PR `X-Total-Count` uses a `COUNT(*)` scoped to `flow_id`, which is acceptable for the current query shape. For future tag-based JSONB filtering, I agree we should revisit with a dedicated strategy (likely GIN index plus possibly different total-count behavior under tag filters). I documented this trade-off in code and kept it out of scope for this PR.

services/metadata_service/api/run.py
GunaPalanivel: I scoped this PR to runs because runs are the unbounded-growth endpoint and the one currently at risk for oversized responses. Steps/tasks have different bounded patterns. I kept the implementation reusable so adding pagination to other endpoints in follow-up PRs is straightforward.

services/metadata_service/api/run.py
GunaPalanivel: You are right. `_limit=0` is not a true backward-compat path unless explicitly sent. I updated behavior so compatibility is preserved when `_limit` is omitted (legacy unbounded path). I removed `_limit=0` opt-out and now require positive `_limit` when pagination is requested; tests were updated accordingly.

Aryan95614:
Nice approach. Using run_number as the cursor key is actually cleaner for ordering stability than ts_epoch since it's a monotonic PK with no tie risk. I went with ts_epoch on the GSoC fork (saikonen/metaflow-service PR #9) because it's available on all six table types (flows, steps, tasks, artifacts, metadata all have ts_epoch but not all have a sequential PK equivalent). Tradeoff is needing a compound cursor for tiebreaking which I raised in Issue #11 on the GSoC fork. Curious if you're planning to extend beyond runs to the other endpoints?

saikonen:
Closing as already completed

PR #467, opened 2026-03-18
fix: prevent SQL injection in update_row by using parameterized queries

ahmedelgyar1:
## Summary
Fix critical SQL injection vulnerability in `AsyncPostgresTable.update_row()` by replacing string concatenation with psycopg2 parameterized queries.

## Context / Motivation
The [update_row](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:362:4-420:50) method constructed SQL WHERE and SET clauses using direct string concatenation, allowing potential SQL injection through user-controlled inputs like `flow_id`, `run_number`, `step_name`, and [task_id](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:139:4-145:59) (which come from URL path parameters). Fixes the SQL injection vulnerability in [update_row()](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:362:4-420:50).

## Changes Made
- **[update_row()](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:362:4-420:50)**: Replaced direct value interpolation with `%s` placeholders and separate value lists (`set_values`, `filter_values`) passed to `cur.execute()`
- **[update_run_tags()](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:630:4-638:45)**: Removed `QuotedString` wrapper — it produced pre-quoted SQL literals that caused double-escaping with the now-parameterized [update_row()](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:362:4-420:50). Values are now passed as plain `json.dumps()` and escaped by psycopg2 automatically
- Column names and operators remain as Python string formatting since they originate from hardcoded server code, not user input

## Testing
- All 207 integration tests pass via `docker-compose -f docker-compose.test.yml`
- Verified all 3 callers of [update_row()](cci:1://file:///home/ahmed-elgyar/metaflow-service/services/data/postgres_async_db.py:362:4-420:50):
  - `AsyncRunTablePostgres.update_heartbeat` — passes server-generated integers ✅
  - `AsyncRunTablePostgres.update_run_tags` — now passes `json.dumps()` directly ✅
  - `AsyncTaskTablePostgres.update_heartbeat` — passes server-generated integers ✅

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -1,6 +1,5 @@
 import psycopg2
 import psycopg2.extras
-from psycopg2.extensions import QuotedString
 import os
 import aiopg
 import json
@@ -361,41 +360,39 @@ class AsyncPostgresTable(object):
             return aiopg_exception_handling(error)
 
     async def update_row(self, filter_dict={}, update_dict={}, cur: aiopg.Cursor = None):
-        # generate where clause
+        # Build a single values list in the same order as SQL placeholders:
+        # UPDATE ... SET col1=%s, col2=%s WHERE col3=%s AND col4=%s
+        values = []
+
+        # generate SET clause (values come first in SQL)
+        sets = []
+        for col_name, col_val in update_dict.items():
+            sets.append('%s = %%s' % col_name)
+            values.append(col_val)
+        set_clause = ", ".join(sets)
+
+        # generate WHERE clause (values come second in SQL)
         filters = []
         for col_name, col_val in filter_dict.items():
             operator = '='
-            v = str(col_val).strip("'")
-            if not v.isnumeric():
-                v = "'" + v + "'"
             find_operator = operator_match.match(col_name)
             if find_operator:
                 col_name = find_operator.group(1)
                 operator = find_operator.group(2)
-                filters.append('(%s IS NULL or %s %s %s)' %
-                               (col_name, col_name, operator, str(v)))
+                filters.append('(%s IS NULL or %s %s %%s)' %
+                               (col_name, col_name, operator))
             else:
-                filters.append(col_name + operator + str(v))
+                filters.append('%s %s %%s' % (col_name, operator))
+            values.append(col_val)
+        where_clause = " and ".join(filters)
 
-        seperator = " and "
-        where_clause = ""
-        if bool(filter_dict):
-            where_clause = seperator.join(filters)
-
-        sets = []
-        for col_name, col_val in update_dict.items():
-            sets.append(col_name + " = " + str(col_val))
-
-        set_seperator = ", "
-        set_clause = ""
-        if bool(filter_dict):
-            set_clause = set_seperator.join(sets)
         update_sql = """
                 UPDATE {0} SET {1} WHERE {2};
         """.format(self.table_name, set_clause, where_clause)
+        values = tuple(values)
 
         async def _execute_update_on_cursor(_cur):
-            await _cur.execute(update_sql)
+            await _cur.execute(update_sql, values)
             if _cur.rowcount < 1:
                 return DBResponse(response_code=404,
                                   body={"msg": "could not find row"})
@@ -631,7 +628,7 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
         filter_dict = {"flow_id": flow_id,
                        run_key: str(run_value)}
 
-        set_dict = {"tags": QuotedString(json.dumps(run_tags)).getquoted().decode()}
+        set_dict = {"tags": json.dumps(run_tags)}
         return await self.update_row(filter_dict=filter_dict,
                                      update_dict=set_dict,
                                      cur=cur)


services/data/postgres_async_db.py
```diff
@@ -631,7 +633,7 @@ async def update_run_tags(self, flow_id: str, run_id: str, run_tags: list, cur:
         filter_dict = {"flow_id": flow_id,
                        run_key: str(run_value)}
 
-        set_dict = {"tags": QuotedString(json.dumps(run_tags)).getquoted().decode()}
+        set_dict = {"tags": json.dumps(run_tags)}
```
saikonen, line 631: QuotedString is an unused import now.

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
 
         set_seperator = ", "
         set_clause = ""
-        if bool(filter_dict):
+        if bool(update_dict):
             set_clause = set_seperator.join(sets)
         update_sql = """
                 UPDATE {0} SET {1} WHERE {2};
         """.format(self.table_name, set_clause, where_clause)
+        values = tuple(set_values + filter_values)
```
saikonen: this is a bit sensitive to ordering now. It might be a good idea to see if we can combine the construction of the two sets of values into one instead to keep the ordering easier to reason about.

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
+        set_values = []
         for col_name, col_val in update_dict.items():
-            sets.append(col_name + " = " + str(col_val))
+            sets.append('%s = %%s' % col_name)
+            set_values.append(col_val)
 
         set_seperator = ", "
         set_clause = ""
-        if bool(filter_dict):
+        if bool(update_dict):
```
saikonen: is the bool check actually necessary? wouldn't the join just produce an empty string with an empty set anyway?

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
+                filters.append('(%s IS NULL or %s %s %%s)' %
+                               (col_name, col_name, operator))
             else:
-                filters.append(col_name + operator + str(v))
+                filters.append('%s %s %%s' % (col_name, operator))
+            filter_values.append(col_val)
 
         seperator = " and "
         where_clause = ""
         if bool(filter_dict):
```
saikonen: same here, though this is historical cleanup at this point. The conditional shouldn't actually be necessary, as the join will produce an empty string with an empty list.

services/data/postgres_async_db.py
ahmedelgyar1: I’ll remove it in the next commit, sorry I missed that

services/data/postgres_async_db.py
ahmedelgyar1: You're right the ordering is sensitive. I see two approaches to solve this :
1-Build a single list from the start, appending values in the same order they appear in the sql set values first, then where values. The code order mirrors the sql order.
2- Store each placeholder alongside its value as a pair, then unzip them at the end. This makes it impossible to accidentally misorder them.
What do you prefer?

services/data/postgres_async_db.py
ahmedelgyar1: You're right it's not necessary. I will remove it

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
     async def update_row(self, filter_dict={}, update_dict={}, cur: aiopg.Cursor = None):
```
snarnawareEntrata: ### 🟡 Parameterized values added but column names and operators remain unsanitized
**Category:** api_contract | **Severity:** medium | **Confidence:** 84%

The change correctly parameterizes filter and update values, but SQL identifiers and operators are still interpolated directly into the SQL string from `filter_dict` and `update_dict` keys. If any caller can influence column names such as `col_name` or operator suffixes matched by `operator_match`, this can still produce malformed SQL or SQL injection through identifiers/operators. This is a request validation/sanitization gap that affects the API surface exposed by these data-layer methods.

**Suggestion:** Restrict `filter_dict` and `update_dict` keys to a fixed allowlist of known column names and allowed operators before constructing SQL. Use `psycopg2.sql.Identifier`/`SQL` composition for identifiers instead of string interpolation.

---
*PR Guardrail Agent — automated review*

snarnawareEntrata: ### 🟡 Unsafe fallback behavior for empty filter/update inputs is not guarded at the abstraction boundary
**Category:** architecture | **Severity:** medium | **Confidence:** 74%

The method uses mutable default dictionaries and builds SQL even when `filter_dict` or `update_dict` may be empty. Although one bug is corrected (`if bool(update_dict)`), the abstraction still appears to permit malformed operations such as `UPDATE table SET  WHERE ...` or `UPDATE table SET ... WHERE ;` depending on caller input. In a low-level shared data-access utility, missing validation is an architecture issue because callers across the codebase inherit unsafe behavior rather than receiving a constrained API.

**Suggestion:** Reject empty `update_dict` and, unless intentionally supported, reject empty `filter_dict` to prevent accidental broad updates. Also replace mutable default arguments with `None` and initialize inside the function. Consider narrowing the API so callers cannot construct invalid update shapes.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
-                filters.append('(%s IS NULL or %s %s %s)' %
-                               (col_name, col_name, operator, str(v)))
+                filters.append('(%s IS NULL or %s %s %%s)' %
+                               (col_name, col_name, operator))
             else:
-                filters.append(col_name + operator + str(v))
+                filters.append('%s %s %%s' % (col_name, operator))
+            filter_values.append(col_val)
 
         seperator = " and "
```
snarnawareEntrata: ### 🟡 Potentially unbounded UPDATE when filter_dict is empty
**Category:** db_query | **Severity:** medium | **Confidence:** 73%

The method allows an empty filter_dict. In that case, where_clause remains an empty string and the generated SQL becomes `UPDATE <table> SET ... WHERE ;`, which is malformed. While this is primarily a correctness bug rather than a performance issue, the broader concern is that there is no guard ensuring updates are scoped by a predicate. If later modified to omit WHERE entirely, this pattern could update every row in the table. The diff does not show callers, so impact is uncertain.

**Suggestion:** Explicitly reject empty filter_dict for update_row unless a full-table update is intentionally supported and separately gated. Raise an exception before constructing SQL when no filters are provided.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
snarnawareEntrata: ### 🟡 Mutable default arguments for filter_dict/update_dict can retain state across calls
**Category:** null_safety | **Severity:** medium | **Confidence:** 91%

The function `update_row` uses `{}` as the default value for both `filter_dict` and `update_dict`. In Python, default mutable arguments are created once at function definition time and then reused across invocations. While this diff does not show direct mutation of these dictionaries inside the function, this is still a defensive-programming gap and can become a source of unexpected shared state if the function is later modified or if callers pass and mutate references in ways that interact with defaults. It is safer to use `None` defaults and initialize inside the function.

**Suggestion:** Change the signature to `async def update_row(self, filter_dict=None, update_dict=None, cur: aiopg.Cursor = None):` and initialize with `filter_dict = filter_dict or {}` and `update_dict = update_dict or {}` at the top of the function.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
         where_clause = ""
         if bool(filter_dict):
             where_clause = seperator.join(filters)
```
snarnawareEntrata: ### 🟠 Empty update_dict now produces invalid SQL with empty SET clause
**Category:** null_safety | **Severity:** high | **Confidence:** 92%

The code now correctly checks `if bool(update_dict):` before building `set_clause`, but it still proceeds to format and execute `UPDATE {table} SET {1} WHERE {2};` even when `update_dict` is empty. In that case, `set_clause` is an empty string and the generated SQL becomes `UPDATE ... SET  WHERE ...`, which will fail at runtime. This is a defensive-programming issue caused by missing validation of required input before use.

**Suggestion:** Validate `update_dict` before building SQL and return an error or raise an exception if it is empty. Example: `if not update_dict: raise ValueError('update_dict must not be empty')`.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
     async def update_row(self, filter_dict={}, update_dict={}, cur: aiopg.Cursor = None):
         # generate where clause
         filters = []
```
snarnawareEntrata, line 375: ### 🟠 Dynamic column/operator interpolation still allows SQL injection via identifiers
**Category:** db_query | **Severity:** high | **Confidence:** 91%

The change correctly parameterizes values passed to the UPDATE statement, which reduces value-based SQL injection risk. However, SQL identifiers and operators are still interpolated directly into SQL from `filter_dict` and `update_dict` keys. `col_name` comes from dictionary keys and is inserted into both the WHERE and SET clauses without validation or quoting. Additionally, an operator may be extracted from `col_name` via `operator_match` and interpolated directly. If any part of these inputs can be influenced by untrusted data, an attacker could inject arbitrary SQL through column names or crafted operator suffixes. Parameter binding does not protect SQL identifiers or operators.

**Suggestion:** Restrict filter and update keys to a fixed allowlist of known column names, and restrict operators to a small explicit allowlist (for example `=`, `!=`, `<`, `>`, `<=`, `>=`). Build SQL using validated identifiers only, ideally with psycopg2.sql.Identifier / SQL composition rather than string formatting.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
+        filter_values = []
         for col_name, col_val in filter_dict.items():
             operator = '='
-            v = str(col_val).strip("'")
-            if not v.isnumeric():
-                v = "'" + v + "'"
             find_operator = operator_match.match(col_name)
             if find_operator:
                 col_name = find_operator.group(1)
                 operator = find_operator.group(2)
```
snarnawareEntrata, line 381: ### 🟠 Empty filter_dict can generate malformed or unintended UPDATE statement
**Category:** null_safety | **Severity:** high | **Confidence:** 90%

If `filter_dict` is empty, `where_clause` remains an empty string, but the SQL still includes `WHERE {2}`. This produces malformed SQL (`... WHERE ;`) or, if later refactored to omit `WHERE`, risks updating all rows. The method does not defensively validate that a filter is present before executing an update. Given historical incidents involving unsafe concurrent updates, missing safeguards around update scope are especially risky.

**Suggestion:** Require a non-empty `filter_dict` for updates unless an explicit full-table update mode is intended. Add a guard such as `if not filter_dict: raise ValueError('filter_dict must not be empty for update_row')`.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
snarnawareEntrata, line 362: ### 🟡 Mutable default arguments for filter_dict/update_dict can leak state across calls
**Category:** null_safety | **Severity:** medium | **Confidence:** 95%

The function signature uses `{}` as default values for `filter_dict` and `update_dict`. In Python, default mutable objects are created once at function definition time and then shared across invocations. While this method does not currently mutate these dictionaries directly in the shown diff, this remains a defensive-programming gap and can become a source of unexpected behavior if future changes mutate them or if callers rely on isolation between calls. It also weakens null-safety because callers may omit arguments and the function will operate on shared mutable state instead of a fresh default.

**Suggestion:** Change the signature to `async def update_row(self, filter_dict=None, update_dict=None, cur: aiopg.Cursor = None):` and initialize with `filter_dict = filter_dict or {}` / `update_dict = update_dict or {}` inside the function.

---
*PR Guardrail Agent — automated review*

snarnawareEntrata, line 362: ### 🟡 Mutable default arguments in data-access API introduce shared state risk
**Category:** architecture | **Severity:** medium | **Confidence:** 94%

The method signature uses `{}` as defaults for `filter_dict` and `update_dict`. In Python, mutable defaults are shared across invocations. Even if this function does not currently mutate them, this is a known footgun and an architectural smell in a shared DB utility because later changes can accidentally create cross-request state leakage. In async services, hidden shared mutable state is especially risky.

**Suggestion:** Change defaults to `None` and initialize inside the function: `filter_dict = filter_dict or {}` and `update_dict = update_dict or {}`. Apply the same rule consistently across shared service-layer helpers.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -363,39 +363,41 @@ async def run_in_transaction_with_serializable_isolation_level(self, fun):
...
-                filters.append(col_name + operator + str(v))
+                filters.append('%s %s %%s' % (col_name, operator))
+            filter_values.append(col_val)
 
         seperator = " and "
         where_clause = ""
         if bool(filter_dict):
             where_clause = seperator.join(filters)
 
         sets = []
```
snarnawareEntrata: ### 🟡 UPDATE can become invalid or unsafe when update_dict is empty
**Category:** db_query | **Severity:** medium | **Confidence:** 93%

The patch fixes a previous bug by checking `update_dict` before building the SET clause, but the function still proceeds even if `update_dict` is empty. In that case `set_clause` remains empty and SQL becomes `UPDATE table SET  WHERE ...`, which will fail at runtime. This is a correctness issue and can also trigger unnecessary transaction retries or connection churn under load.

**Suggestion:** Validate that `update_dict` is non-empty before generating SQL, and return a clear error if no columns are provided to update.

---
*PR Guardrail Agent — automated review*

ahmedelgyar1:
Hi @saikonen  I’ve implemented the first approach .
Could you please take a look and let me know your feedback?
Also, if you think the second approach would be better, I can switch to it.

saikonen (changes requested):
changes look good. I've fixed the issue with tests not executing on `master` so could you rebase this PR and cover the linter errors. This should be good to go otherwise.

ahmedelgyar1:
> changes look good. I've fixed the issue with tests not executing on `master` so could you rebase this PR and cover the linter errors. This should be good to go otherwise.

@saikonen I’ve rebased the PR on the latest master and addressed the linting/test issues. Could you please take another look?

saikonen (approved):

PR #471, opened 2026-03-24
feat: add opt-in DBQueryTracer for per-request DB query tracing

Youssef-SH:
## Summary

Implements Issue [#5](https://github.com/valayDave/metaflow-service/issues/5) by adding opt-in per-request DB query tracing in `execute_sql()`.

All DB access flows through this method. This introduces `DBQueryTracer` to record:

- query count  
- rows returned  
- execution time  

Tracing is scoped via `ContextVar`, ensuring isolation per async request.  
When disabled, it is a no-op (single conditional check, no overhead).

## Validation

Tested on the dev stack using real endpoints:

**GET /flows/TestFlow/runs**

- query_count: 1  
- total_rows: 3  

**GET /flows/TestFlow/runs/3/steps/start/tasks**

- query_count: 2  
- total_rows: 2  

The first endpoint executes a single DB query.  
The second demonstrates query fan-out across multiple DB calls.

## Notes

- Fully orthogonal to API behavior and schema  
- Single instrumentation point (`execute_sql()`)  
- Enables per-request DB cost visibility for agent workloads

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -56,7 +56,8 @@ class DBQueryTracer:
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        active_db_tracer.reset(self._tokens.pop())
+        if self._tokens:
+            active_db_tracer.reset(self._tokens.pop())
 
     def _record(self, execution_time, row_count):
         self._queries.append(
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -7,6 +7,7 @@ import json
 import math
 import re
 import time
+from contextvars import ContextVar
 from services.utils import logging, DBType
 from typing import List, Tuple
 
@@ -37,10 +38,40 @@ ARTIFACT_TABLE_NAME = os.environ.get("DB_TABLE_NAME_ARTIFACT", "artifact_v3")
 DB_SCHEMA_NAME = os.environ.get("DB_SCHEMA_NAME", "public")
 
 operator_match = re.compile('([^:]*):([=><]+)$')
+active_db_tracer = ContextVar("active_db_tracer", default=None)
 
 # use a ddmmyyy timestamp as the version for triggers
 TRIGGER_VERSION = "05092024"
 TRIGGER_NAME_PREFIX = "notify_ui"
+
+
+# Tracing is scoped via ContextVar, ensuring metrics are isolated per request context
+class DBQueryTracer:
+    def __init__(self):
+        self._queries = []
+        self._tokens = []
+
+    def __enter__(self):
+        self._tokens.append(active_db_tracer.set(self))
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        active_db_tracer.reset(self._tokens.pop())
+
+    def _record(self, execution_time, row_count):
+        self._queries.append(
+            {"execution_time": execution_time, "row_count": row_count}
+        )
+
+    def summary(self):
+        query_count = len(self._queries)
+        total_rows = sum(query["row_count"] for query in self._queries)
+        total_time = sum(query["execution_time"] for query in self._queries)
+        return {
+            "query_count": query_count,
+            "total_time": total_time,
+            "total_rows": total_rows,
+        }
 
 
 class _AsyncPostgresDB(object):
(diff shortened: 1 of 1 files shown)


services/data/postgres_async_db.py
```diff
@@ -249,27 +280,46 @@ async def execute_sql(self, select_sql: str, values=[], fetch_single=False,
                           expanded=False, limit: int = 0, offset: int = 0,
```
snarnawareEntrata, line 281: ### 🟡 Mutable default argument for SQL parameter values
**Category:** null_safety | **Severity:** medium | **Confidence:** 88%

The `execute_sql` function uses `values=[]` as a default parameter. Default mutable arguments are shared across invocations, which is a defensive programming risk and can lead to unexpected state leakage if the list is ever mutated before being passed to `_cur.execute`. While this diff did not introduce the issue, it remains present in the affected function and is relevant to null/undefined safety because shared mutable defaults can create hard-to-diagnose invalid states.

**Suggestion:** Change the function signature to use `values=None` and initialize it inside the function, e.g. `if values is None: values = []`.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -37,12 +38,42 @@
 DB_SCHEMA_NAME = os.environ.get("DB_SCHEMA_NAME", "public")
 
 operator_match = re.compile('([^:]*):([=><]+)$')
```
snarnawareEntrata, line 40: ### 🟡 Request-scoped tracing implemented via module-level implicit context
**Category:** architecture | **Severity:** medium | **Confidence:** 76%

The new tracing facility is wired through a module-level `ContextVar` and accessed implicitly inside `execute_sql`. While `ContextVar` provides async-context isolation, this still introduces hidden ambient state into the data access layer. Callers and tests must know to set global context before invoking DB methods, which weakens abstraction boundaries and makes dependencies less explicit than passing a tracer/metrics collector through the call chain or injecting instrumentation at the DB client boundary.

**Suggestion:** Prefer explicit instrumentation injection: pass an optional tracer/metrics interface into the DB class or `execute_sql`, or wrap the DB adapter with a decorator/interceptor responsible for tracing. If `ContextVar` is retained, isolate it behind a small instrumentation abstraction rather than exposing module-global ambient state directly.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
snarnawareEntrata, line 281: ### 🟡 Instrumentation concerns are mixed directly into query execution logic
**Category:** architecture | **Severity:** medium | **Confidence:** 74%

The `execute_sql` path now handles SQL execution, result materialization, pagination, and tracing/timing/error metric recording. This increases the responsibility of an already central data access method and couples observability concerns to persistence logic. Over time, adding more metrics or tracing fields here can turn this method into a broader god-function hotspot.

**Suggestion:** Extract tracing into a helper/decorator/interceptor around cursor execution, e.g. a `_timed_execute_fetch(...)` helper or a generic DB instrumentation wrapper. Keep `execute_sql` focused on query semantics and response shaping while observability is handled by a dedicated abstraction.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
```diff
@@ -37,12 +38,42 @@
...
 # use a ddmmyyy timestamp as the version for triggers
 TRIGGER_VERSION = "05092024"
 TRIGGER_NAME_PREFIX = "notify_ui"
 
 
+# Tracing is scoped via ContextVar, ensuring metrics are isolated per request context
+class DBQueryTracer:
+    def __init__(self):
+        self._queries = []
+        self._tokens = []
```
snarnawareEntrata, line 52: ### 🟡 ContextVar reset may pop from empty token stack
**Category:** null_safety | **Severity:** medium | **Confidence:** 73%

`DBQueryTracer.__exit__` unconditionally calls `self._tokens.pop()` before resetting the `ContextVar`. If `__exit__` is invoked without a matching successful `__enter__`, or if token stack balance is disrupted by misuse/reentrancy, this will raise `IndexError`. This is not a null dereference exactly, but it is an unguarded access to an element that may not exist and can crash at cleanup time.

**Suggestion:** Defensively guard the pop operation, e.g. `if self._tokens: active_db_tracer.reset(self._tokens.pop())`, or enforce and validate correct lifecycle usage with a clearer error message.

---
*PR Guardrail Agent — automated review*

services/data/postgres_async_db.py
Youssef-SH: Using ContextVar intentionally to keep tracing request-scoped without changing existing method signatures or call sites. Keeps it non-invasive and opt-in.

services/data/postgres_async_db.py
Youssef-SH: Good catch, added a guard to avoid potential empty stack pop.

services/data/postgres_async_db.py
Youssef-SH: Not introduced in this PR. Agree it should be fixed separately to avoid scope creep.

Youssef-SH: Keeping tracing inline to preserve a single instrumentation point in execute_sql. Can be refactored later if needed.

PR #472, opened 2026-03-26
Add limit cap to prevent large payload in runs API

rachit018:
> This PR adds a limit cap to the get_all_runs endpoint to prevent excessively large payloads.

Currently, users can request a very large number of runs which may impact performance. This change ensures that the maximum limit is capped at 100.

This improves API stability and prevents potential performance issues.


rachit018:
Hi, I’ve opened this PR to add a limit cap to the runs API to prevent large payloads.

I’d appreciate any feedback or suggestions. Thanks!

PR #481, opened 2026-06-06
Add filtering to the runs endpoint (status, time, user, tags) with cursor support

Aryan95614:
## What

Adds server-side filtering to `GET /flows/{flow_id}/runs`, composed with the cursor pagination from #482.

Filters use the ui_backend `field:operator` grammar (per the earlier review on the time-range work):

- `status:eq=running` (or `status:eq=running,failed` for OR)
- `ts_epoch:ge=123&ts_epoch:le=456` (inclusive epoch-millisecond bounds)
- `user:eq=alice` (verified owner: the run must carry a `user:<name>` system tag, matching what the UI labels as the owner)
- `_tags:any=a,b` (OR) / `_tags:all=a,b` (AND)

The parsers are lifted verbatim from ui_backend into a new web-free `services/data/filter_grammar.py`, so both services parse filters identically from one definition rather than a copy.

## Cursor pagination

Filtering composes with the cursor pagination added in #482. `get_filtered_runs_paginated` walks the filtered result set by the same `(ts_epoch, run_number)` keyset, fetches `limit + 1` to detect a next page, and returns the `X-Next-Cursor` header. With no `_limit`/`_cursor` the endpoint keeps its raw-array response over the filtered set.

## Notes

- `status` is the only derived column needing the lateral joins, so they are gated behind a status filter; `user` is a join-free derived column.
- `status` is derived relative to the heartbeat cutoff, so a multi-status filter that includes `running` is time-relative, not fully deterministic. This is documented on the endpoint.
- All filter values are bound as query parameters, never interpolated. A bad status value, operator, time bound, or cursor returns 400.

This consolidates the time-range and tag filtering work (previously separate PRs on the fork) into the shared grammar.

Verified against Postgres: full integration suite green (133 passed, 13 skipped), including the cursor pagination test and a filtering-plus-cursor composition test.

### DIFF
--- a/services/data/models.py
+++ b/services/data/models.py
@@ -43,6 +43,8 @@ class RunRow(object):
         tags=None,
         system_tags=None,
         last_heartbeat_ts=None,
+        status=None,
+        user=None,
     ):
         self.flow_id = flow_id
         self.user_name = user_name
@@ -55,10 +57,14 @@ class RunRow(object):
 
         self.ts_epoch = ts_epoch
         self.last_heartbeat_ts = last_heartbeat_ts
+        # derived, only present when the row was fetched with the status join
+        self.status = status
+        # derived verified owner, present when fetched with the filter select columns
+        self.user = user
 
     def serialize(self, expanded: bool = False):
         if expanded:
-            return {
+            body = {
                 "flow_id": self.flow_id,
                 "run_number": self.run_number,
                 "run_id": self.run_id,
@@ -69,7 +75,7 @@ class RunRow(object):
                 "last_heartbeat_ts": self.last_heartbeat_ts,
             }
         else:
-            return {
+            body = {
                 "flow_id": self.flow_id,
                 "run_number": get_exposed_run_id(self.run_number, self.run_id),
                 "user_name": self.user_name,
@@ -78,6 +84,11 @@ class RunRow(object):
                 "system_tags": self.system_tags,
                 "last_heartbeat_ts": self.last_heartbeat_ts,
             }
+        if self.status is not None:
+            body["status"] = self.status
+        if self.user is not None:
+            body["user"] = self.user
+        return body
 
 
 class StepRow(object):
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -784,38 +784,6 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
 
         return await self.get_records(filter_dict=filter_dict)
 
-    async def get_filtered_runs(
-        self,
-        conditions: List[str],
-        values: list,
-        enable_joins: bool = False,
-        limit: int = 0,
-        offset: int = 0,
-        order: List[str] = None,
-    ):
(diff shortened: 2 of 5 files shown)


services/data/postgres_async_db.py
```diff
@@ -583,9 +588,74 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
...
     trigger_keys = primary_keys + ["last_heartbeat_ts"]
-    select_columns = keys
     flow_table_name = AsyncFlowTablePostgres.table_name
 
+    # Derived run status (running/completed/failed). This mirrors the definition in
+    # ui_backend_service so filtering agrees with what the UI shows. The two lateral
+    # joins below pull the 'end' step's attempt metadata that the status depends on.
+    # Only used when joins are enabled (i.e. when filtering by status), so the plain
+    # get_all_runs path stays a simple, join-free query.
+    joins = [
```
saikonen, line 685: battle-tested query logic from the ui_backend side, and this is gated behind the status query param, so from a performance perspective it should be fine to add to the metadata-service as well.

The only concern is with `status=failed&status=running` which is in the case of relying on heartbeats, not entirely deterministic.

saikonen (commented):
looks good so far. Let's wait for the run pagination PR to go in so this can be modified to support cursor based pagination as well.

services/data/postgres_async_db.py
```diff
@@ -742,22 +784,79 @@ async def get_all_runs(self, flow_id: str):
 
         return await self.get_records(filter_dict=filter_dict)
 
-    async def get_all_runs_paginated(
-        self, flow_id: str, cur_ts: int = None, cur_run: int = None, limit: int = None
+    async def get_filtered_runs(
```
saikonen: this one is probably unnecessary, as the new clients will support both filtering and cursor pagination. We do need to keep the unfiltered-unpaginated fetch for backwards compatibility though.

services/metadata_service/api/run.py
```diff
@@ -108,12 +216,19 @@ async def get_all_runs(self, request):
                 return DBResponse(response_code=400, body="Invalid cursor")
 
         if limit is None and cursor is None:
-            return await self._async_table.get_all_runs(flow_name)
+            return await self._async_table.get_filtered_runs(
+                conditions=conditions, values=values, enable_joins=enable_joins
+            )
```
saikonen: This should fallback to legacy method with no pagination or filters

services/data/models.py
```diff
@@ -78,6 +84,11 @@ def serialize(self, expanded: bool = False):
                 "system_tags": self.system_tags,
                 "last_heartbeat_ts": self.last_heartbeat_ts,
             }
+        if self.status is not None:
+            body["status"] = self.status
+        if self.user is not None:
+            body["user"] = self.user
+        return body
```
saikonen, line 91: Existing clients might handle unsupported extra fields gracefully, but would need to verify that.

This is a good safety measure with that in mind though 👌

saikonen (approved):

PR #482, opened 2026-06-15
Add cursor-based pagination to GET /flows/{flow_id}/runs

guts27:
Adds `_limit` and `_cursor` query parameters to `GET /flows/{flow_id}/runs`. This allows clients to get runs page by page, instead of getting all rows at once.

The cursor is a opaque base64 encoded string of (ts_epoch, run_number). Clients can use the X-Next-Cursor response header to get the next page. If a client doesn't send these parameters, the endpoint returns all runs like before, so it doesn't break existing clients. Also, the cursor is in the header, not the body, so the response structure stays exactly the same.

The default _limit is 50, and the maximum is 500. Results are ordered by ts_epoch DESC, run_number DESC (newest first). The query uses keyset pagination, so it finds the next page using the index without OFFSET. An invalid _cursor returns a 400 error.

Before this PR, the endpoint returned all runs in one response. This is slow and uses too much memory if a flow has many runs. The cursor approach only reads one page (e.g. 50 rows) no matter how big the flow is:

**Before (full scan)**

| Scale | Time | Memory |
|-------|------|--------|
| 1K | 8 ms | — |
| 100K | 1012 ms | — |
| 1M | 4459 ms | 190 MB |

**After (cursor, 50 per page)**

| Scale | Time | Rows read |
|-------|------|-----------|
| 1K | 0.07 ms | 50 |
| 100K | 4.3 ms | 50 |
| 1M | 26 ms | 50 |

I added a composite index (flow_id, ts_epoch DESC, run_number DESC) and a migration file. This makes the cursor query use an index scan instead of a sequential scan. EXPLAIN ANALYZE on 1M rows shows:

Limit  (cost=0.43..1.30 rows=6 width=646) (actual time=0.756..0.779 rows=6 loops=1)
  ->  Index Scan using runs_v3_idx_flow_ts_runnum_desc on runs_v3  (cost=0.43..145103.95 rows=998561 width=646) (actual time=0.755..0.758 rows=6 loops=1)
        Index Cond: (((flow_id)::text = 'test_1m'::text) AND (ROW(ts_epoch, run_number) < ROW('1780155056177'::bigint, 101005)))
Planning Time: 40.171 ms
Execution Time: 3.144 ms

Tested with real Postgres: no parameters (returns all runs), _limit only (first page), fetching the next page with cursor, reaching the last page (no header), invalid cursor (400), and a flow that doesn't exist (200 with an empty array).

Question for reviewers (Known limitation):
When a run has a run_id, the get_exposed_run_id function replaces run_number with the string run_id during serialization. So the cursor might encode a string instead of an integer. This PR works fine assuming run_id is NULL.

Ideally, the cursor should use the raw integer run_number before serialization. However, the cursor is currently created using the already-serialized response.body. The raw run_number is hidden inside the execute_sql utility function, so it is hard to extract it cleanly. I'd appreciate your input on the right direction here.


guts27:
I pushed the `run_id` fix we discussed.

**What I changed:**
I didn't add a new flag to `find_records`. Instead, I added a `run_number` field to `DBPagination`. I extract this raw integer inside `execute_sql` before serialization. Now, the cursor encodes this raw integer directly. It works correctly even if a run has an explicit `run_id`.

**Why I chose this way:**

- It keeps the function signatures (`find_records`, `execute_sql`) the same.
- The query stays single-shot.
- The new field defaults to `None`, so it doesn't break other endpoints.

**Testing:**
I tested it with Postgres. I set a `run_id` on the first few runs.

- The cursor correctly encodes the integer , not the string
- Pagination works perfectly across those runs.
- Empty flows, invalid cursors, and old clients still work like before.

If you prefer the flag approach, please let me know. I can easily change it.

*Note:* I tested all of this manually for now. I will add automated integration tests in a follow-up PR. But if you want the tests in this PR before merging, just let me know.

user@USER:~/projects/metaflow-service$ curl -i "http://localhost:8080/flows/test_1k/runs?_limit=5"
HTTP/1.1 200 OK
Content-Type: application/json
METADATA_SERVICE_VERSION: 2.5.0
X-Limit: 5
X-Next-Cursor: eyJ0c19lcG9jaCI6IDE3ODAxNTUwNTYxNzcsICJydW5fbnVtYmVyIjogNX0=
Content-Length: 1253
Date: Wed, 17 Jun 2026 13:19:40 GMT
Server: Python/3.11 aiohttp/3.14.1

[{"flow_id": "test_1k", "run_number": "custom-run-1", "user_name": "eunhye", "ts_epoch": 1780155060177, "tags": ["tag:retried", "env:prod", "env:dev"], "system_tags": ["user:eunhye", "metaflow_version:2.11.0", "metaflow_version:2.11.0"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-2", "user_name": "sakari", "ts_epoch": 1780155059177, "tags": ["env:dev"], "system_tags": ["user:sakari", "python_version:3.11", "runtime:kubernetes"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-3", "user_name": "sakari", "ts_epoch": 1780155058177, "tags": ["tag:failed", "env:prod", "tag:retried"], "system_tags": ["user:sakari", "runtime:local", "metaflow_version:2.11.0"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-4", "user_name": "aryan", "ts_epoch": 1780155057177, "tags": ["tag:failed", "tag:retried"], "system_tags": ["user:aryan", "python_version:3.10", "metaflow_version:2.11.0"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-5", "user_name": "sakari", "ts_epoch": 1780155056177, "tags": ["tag:success", "tag:failed"], "system_tags": ["user:sakari", "runtime:kubernetes", "metaflow_version:2.10.0"], "last_heartbeat_ts": null}]

user@USER:~/projCURSOR=$(curl -s -i "http://localhost:8080/flows/test_1k/runs?_limit=5" | grep -i "x-next-cursor:" | awk '{print $2}' | tr -d '\r')rint $2}' | tr -d '\r')

user@USER:~/projects/metaflow-service$ curl -i "http://localhost:8080/flows/test_1k/runs?_limit=5&_cursor=$CURSOR"
HTTP/1.1 200 OK
Content-Type: application/json
METADATA_SERVICE_VERSION: 2.5.0
X-Limit: 5
X-Next-Cursor: eyJ0c19lcG9jaCI6IDE3ODAxNTUwNTExNzcsICJydW5fbnVtYmVyIjogMTB9
Content-Length: 1273
Date: Wed, 17 Jun 2026 13:19:58 GMT
Server: Python/3.11 aiohttp/3.14.1

[{"flow_id": "test_1k", "run_number": "custom-run-6", "user_name": "aryan", "ts_epoch": 1780155055177, "tags": ["tag:success", "env:dev"], "system_tags": ["user:aryan", "metaflow_version:2.11.0", "runtime:local"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-7", "user_name": "sakari", "ts_epoch": 1780155054177, "tags": ["tag:failed", "tag:retried", "env:dev"], "system_tags": ["user:sakari", "python_version:3.11", "metaflow_version:2.10.0"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-8", "user_name": "aryan", "ts_epoch": 1780155053177, "tags": ["tag:retried", "env:dev", "tag:failed"], "system_tags": ["user:aryan", "runtime:local", "runtime:kubernetes"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-9", "user_name": "sakari", "ts_epoch": 1780155052177, "tags": ["tag:success", "env:prod"], "system_tags": ["user:sakari", "python_version:3.11", "metaflow_version:2.10.0"], "last_heartbeat_ts": null}, {"flow_id": "test_1k", "run_number": "custom-run-10", "user_name": "eunhye", "ts_epoch": 1780155051177, "tags": ["env:prod", "tag:retried", "tag:failed"], "system_tags": ["user:eunhye", "metaflow_version:2.10.0", "metaflow_version:2.11.0"], "last_heartbeat_ts": null}]

services/data/db_utils.py
```diff
@@ -111,3 +113,35 @@ def filter_artifacts_by_attempt_id_for_tasks(
...
+            decoded = json.loads(b64decode(cursor).decode())
+
+        except (binascii.Error, UnicodeDecodeError,json.JSONDecodeError ):
+            raise ValueError("invalid_cursor")
+
+        if "ts_epoch" not in decoded or "run_number" not in decoded:
+            raise ValueError("invalid_cursor")
+        return decoded  
+            
+    return None
```
saikonen: is the `None` return valid as a default? when would `decode_cursor` be called with a `None` instead of a proper cursor value?

services/data/db_utils.py
```diff
@@ -111,3 +113,35 @@ def filter_artifacts_by_attempt_id_for_tasks(
...
+
+def encode_cursor(
+    cursor:dict,
+    encoder: Callable[[bytes], str] = default_encoder, 
+) -> str | None:
+    if cursor:
+        cursor = json.dumps(cursor).encode() if isinstance(cursor, dict ) else cursor
+        encoded = encoder(cursor)
+
+        return encoded
```
saikonen: we can probably just inline the `b64encode` in here, as I don't see a need for using any other encoding for the cursors.

saikonen (commented):
apart from the small refactorings, can you also

- [x] add integration tests for the paginated runs endpoint
- [x] cleanup debugging prints and unnecessary comments.

services/metadata_service/api/utils.py
```diff
@@ -8,9 +8,13 @@
 
 from services.utils import get_traceback_str
 
+from services.data.db_utils import DBPagination
+
 version = metadata.version("metadata_service")
 METADATA_SERVICE_VERSION = version
 METADATA_SERVICE_HEADER = 'METADATA_SERVICE_VERSION'
+X_LIMIT = 'X-Limit'
```
saikonen: only used once so this could just be inlined, same as `X-Next-Cursor` is already.

services/migration_service/migration_files/20260615000000_add_runs_cursor_composite_index.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose NO TRANSACTION
+-- +goose Up
```
saikonen, line 2: for the test failures, I think the tests are using the migration service tooling and not just the run_goose.py, so you should update the pointer to the latest version in `services/migration_service/api/__init__.py#version_dict` as well

services/data/db_utils.py
guts27: Right, run.py only calls decode_cursor inside an if cursor: block, so None never reaches it. Removed the None handling and simplified the signature.

services/data/db_utils.py
guts27: Agreed, inlined b64encode directly and removed the pluggable encoder and the Callable import.

services/metadata_service/api/utils.py
guts27: Done, inlined it as 'X-Limit' to match X-Next-Cursor.

services/migration_service/migration_files/20260615000000_add_runs_cursor_composite_index.sql
guts27: Thanks for the pointer — that was it. Registered the new migration version in version_dict.

guts27:
Added integration tests for the paginated endpoint (test_runs_pagination_get) covering first page, cursor continuation, last page, invalid cursor (400). Also removed the debug prints and unnecessary comments.

services/data/db_utils.py
```diff
@@ -1,15 +1,17 @@
 import asyncio
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Callable
```
saikonen: `Callable` not used in this file anymore

services/data/postgres_async_db.py
```diff
@@ -265,11 +296,14 @@ async def _execute_on_cursor(_cur):
...
             # Will raise IndexError in case fetch_single=True and there's no results
             body = rows[0] if fetch_single else rows
+
             pagination = DBPagination(
                 limit=limit,
                 offset=offset,
                 count=count,
                 page=math.floor(int(offset) / max(int(limit), 1)) + 1,
+                # Used for cursor when has_next (records == limit + 1); ignored otherwise
+                run_number=records[-2]['run_number'] if records is not None and len(records) > 1 else None
```
saikonen: Good for the initial PR as we're focused on run pagination, but needs refactoring in the future because run_number will not work as a tie-breaker for paginating tasks for example.

services/data/postgres_async_db.py
```diff
@@ -214,6 +214,37 @@ async def get_records(self, filter_dict={}, fetch_single=False,
         )
         return response
 
+    async def get_records_paginated(self, filter_dict={}, fetch_single=False,
+                                    ordering: List[str] = None, limit: int = 50, expanded=False,
+                                    cur: aiopg.Cursor = None) -> Tuple[DBResponse, DBPagination]:
+        conditions = []
+        values = []
+        for col_name, col_val in filter_dict.items():
+            if "ts_epoch" in col_name:
```
saikonen, line 260: something to consider here is that `ts_epoch` is also going to be a proper user-set filter later, so this needs a slight rework as it would be part of the filter set with requests like 
```
/runs?ts_epoch:gt=1234&ts_epoch:lt=987
```

One option would be to keep the cursor columns and values separate in this function, so they can be handled outside of regular filter logic.

saikonen: Ok to refactor later as well, unless you want to tackle this in the initial PR already.

services/metadata_service/tests/integration_tests/utils.py
```diff
@@ -246,6 +246,56 @@ def _sort_key(r):
...
+    """
+    response = await cli.get(path, params = params or {})
+
+    assert response.status == status
+
+    body = json.loads(await response.text()) if status == 200 else None
+
+    if data is None:
+        return
+    if data_is_unordered_list_of_dicts:
```
saikonen: is the data ever unordered? Don't we explicitly have an order for all paginated endpoints that this assert helper would be used for?

services/data/db_utils.py
guts27: Thank you, i removed it.

services/data/postgres_async_db.py
guts27: Generalized this — instead of pulling run_number directly, it now passes the full record (next_cursor_record) so the cursor column can be chosen per endpoint.

services/metadata_service/tests/integration_tests/utils.py
guts27: You're right — paginated results have a fixed order, so I dropped the unordered option and assert the order directly.

services/data/postgres_async_db.py
guts27: This overlaps with Aryan's filter work, and he suggested we coordinate on it together. I'm happy either way — whether we handle it here or as a follow-up. Let me know what works best for you.

saikonen (approved):

PR #485, opened 2026-06-25
fix: show run as 'running' when task heartbeats active but run heartbeat expired

npow:
## Problem

In foreach flows, when a transient failure causes the orchestrator to stop updating the run-level heartbeat but individual tasks continue executing, the Metaflow UI shows the run as `failed` even though tasks are still running.

Concrete example: `CoplayBucketBackfillFlow` run with 226 tasks — 1 task hit a transient DNS error (`SparkGenieStderrUnknownHost`), orchestrator heartbeat lapsed, but 30 tasks were still running. UI showed `failed` and the `running` filter returned nothing.

## Root Cause

The run `status` SQL in `run.py` falls to `ELSE 'failed'` when the run-level `last_heartbeat_ts` has exceeded `RUN_INACTIVE_CUTOFF_TIME` (6 min default), regardless of whether individual tasks are still heartbeating.

## Fix

Add an `EXISTS` fallback that checks task-level heartbeats (using `HEARTBEAT_THRESHOLD`) before the final `ELSE 'failed'`. Both `status` and `finished_at` are updated consistently:

- `status`: returns `'running'` if any task has a heartbeat within `HEARTBEAT_THRESHOLD`
- `finished_at`: returns `NULL` (not yet finished) in the same condition

Also adds a new integration test covering this scenario.

## Test

New test: `test_run_status_running_when_task_heartbeat_active_and_run_heartbeat_expired`

- Creates a run with an expired run-level heartbeat (7 min old, beyond the 6-min cutoff)
- Adds a task with a fresh heartbeat
- Asserts `status == 'running'` and `finished_at is None`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/services/ui_backend_service/data/db/tables/run.py
+++ b/services/ui_backend_service/data/db/tables/run.py
@@ -111,7 +111,9 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
             WHEN {table_name}.last_heartbeat_ts IS NOT NULL
                 AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)<={heartbeat_cutoff}
             THEN NULL
-            WHEN latest_task_heartbeat.last_heartbeat_ts IS NOT NULL
+            WHEN {table_name}.last_heartbeat_ts IS NOT NULL
+                AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)<={run_stale_threshold}
+                AND latest_task_heartbeat.last_heartbeat_ts IS NOT NULL
                 AND @(extract(epoch from now())-latest_task_heartbeat.last_heartbeat_ts)<={heartbeat_cutoff}
             THEN NULL
             ELSE {table_name}.last_heartbeat_ts*1000
@@ -119,6 +121,7 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
         """.format(
             table_name=table_name,
             heartbeat_cutoff=RUN_INACTIVE_CUTOFF_TIME,
+            run_stale_threshold=RUN_INACTIVE_CUTOFF_TIME * 10,
         ),
         """
         (CASE
@@ -132,7 +135,9 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
             WHEN {table_name}.last_heartbeat_ts IS NOT NULL
                 AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)<={heartbeat_cutoff}
             THEN 'running'
-            WHEN latest_task_heartbeat.last_heartbeat_ts IS NOT NULL
+            WHEN {table_name}.last_heartbeat_ts IS NOT NULL
+                AND @(extract(epoch from now())-{table_name}.last_heartbeat_ts)<={run_stale_threshold}
+                AND latest_task_heartbeat.last_heartbeat_ts IS NOT NULL
                 AND @(extract(epoch from now())-latest_task_heartbeat.last_heartbeat_ts)<={heartbeat_cutoff}
             THEN 'running'
             ELSE 'failed'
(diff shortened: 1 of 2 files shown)


npow:
Addressed all three points — force-pushed.

**1. Performance (correlated subquery → lateral join)**

Switched from an inline correlated `EXISTS` to a `LEFT JOIN LATERAL` that pre-computes `max(last_heartbeat_ts)` across all tasks for each run. The join result is computed once per row and referenced from both the `status` and `finished_at` CASE expressions — no double-scan. Confirmed that `get_all_runs` uses `enable_joins=True`, so this path does run on list queries; the lateral join is the right pattern (matches the existing `end_attempt_ok` / `end_attempt` joins).

An index on `(flow_id, run_number, last_heartbeat_ts)` in the task table would make this optimal for deployments with large foreach histories. That's a follow-up.

**2. Threshold (`HEARTBEAT_THRESHOLD` → `RUN_INACTIVE_CUTOFF_TIME`)**

Changed to `RUN_INACTIVE_CUTOFF_TIME` (6 min default) throughout — the task-level check now uses the exact same window as the run-level check. `HEARTBEAT_THRESHOLD` (60s) was too narrow and would have produced false negatives for slow-heartbeating tasks. Both checks are now symmetric.

**3. Skipped test**

The skipped test (`test_run_status_failed_with_heartbeat_expired_and_failed_task`) predates this PR and tests the inverse scenario: run heartbeat still alive, task heartbeat expired → 'failed'. That behavior was removed in a prior refactor (hence the skip). My change doesn't affect or worsen that path. Filing a follow-up to either restore or formally retire that behavior.

services/ui_backend_service/data/db/tables/run.py
```diff
@@ -69,6 +69,14 @@ class AsyncRunTablePostgres(AsyncPostgresTable):
...
         ) as end_attempt ON true
         """.format(table_name=table_name, metadata_table=metadata_table),
+        """
+        LEFT JOIN LATERAL (
+            SELECT max(last_heartbeat_ts) as last_heartbeat_ts
+            FROM {task_table} t
+            WHERE t.flow_id = {table_name}.flow_id
+            AND t.run_number = {table_name}.run_number
+        ) as latest_task_heartbeat ON true
+        """.format(table_name=table_name, task_table=task_table),
```
saikonen, line 79: does this perform well at scale for the runs listing?

saikonen (commented):
No immediate objections to the change, only some performance concerns. If the queries work fine in your environments with a bunch of runs, then it should be fine.

I'm wondering about the premise for the change though, where a run level heartbeat goes stale, but tasks are still running. How common is this? To my understanding, task and run heartbeats update in lockstep from the same source

services/ui_backend_service/data/db/tables/run.py
Aryan95614: I measured this locally on PostgreSQL 11 with 50,000 runs and 1,000,000 tasks (20 per run), using warm-cache EXPLAIN ANALYZE with timing overhead disabled. The current max lookup and task index took 0.190 ms for a 50-row listing and 120.791 ms for a status scan across all 50,000 runs. A partial index on (flow_id, run_number, last_heartbeat_ts DESC) for non-null heartbeats, paired with ORDER BY last_heartbeat_ts DESC LIMIT 1, reduced those to 0.076 ms and 63.821 ms. The ordinary paged listing is already cheap because the outer scan stops at the limit; the index mainly protects status-filter and high-scan cases. I pushed the working version at https://github.com/Aryan95614/metaflow-service/commit/9b12cc6. It also moves the task-heartbeat rule into the shared run-status definition used by both ui_backend and metadata-service filtering, so the two services do not classify the same run differently. Full integration result: 135 passed, 13 skipped.

PR #487, opened 2026-07-07
Add filtering to the tasks endpoint

Aryan95614:
## Summary
Extends the `field:operator` filtering grammar introduced for the runs endpoint (#481) to `GET /flows/{flow_id}/runs/{run_number}/steps/{step_name}/tasks`. Supported filters (`ts_epoch`, `user_name`, `task_name`, `last_heartbeat_ts`, ...) now narrow the task listing server-side. With no filters supplied the endpoint keeps its original behavior, so the change is backward compatible.

## Design notes
- **Status is intentionally not filterable** on tasks — task status is not derived by the metadata service.
- **Tag filters are excluded** — a task's stored tags are not canonical; the run's tags are grafted on at read time.
- Cursor pagination composes with filtered task queries the same way it does on the runs endpoint.
- Incoming filters are validated against `TASK_ALLOWED_FILTERS`, and every allowed field is projected by the task table, so a valid filter cannot reference a non-projected column.

## Testing
- New integration test `test_tasks_get_filtering` (time-range + user filtering and the backward-compatible unfiltered path), alongside the existing `filtered_tasks` metadata cases.
- Local run via the test compose: metadata task tests **30 passed**; full suite **226 passed, 13 skipped**.

## CI note
The integration job may report one unrelated failure: `services/ui_backend_service/tests/integration_tests/tasks_test.py::test_task_attempts_with_attempt_metadata`. This is a pre-existing timing flake — `pytest.approx(0, rel=...)` degenerates to a near-exact-zero tolerance, so any real elapsed duration fails. It is confined to the UI backend and does not touch this change; it is fixed separately on `fix-flaky-attempt-duration-tolerance`. Rerun in isolation 5x: 4 passed / 1 failed, confirming non-determinism.

### DIFF
--- a/services/metadata_service/api/task.py
+++ b/services/metadata_service/api/task.py
@@ -1,11 +1,63 @@
 from services.data import TaskRow
 from services.data.postgres_async_db import AsyncPostgresDB
 from services.data.tagging_utils import apply_run_tags_to_db_response
+from services.data.db_utils import DBResponse, translate_run_key
+from services.data.filter_grammar import custom_conditions_query_dict, operators_to_sql
 from services.utils import has_heartbeat_capable_version_tag, read_body
 from services.metadata_service.api.utils import format_response, handle_exceptions
 import json
 from aiohttp import web
 import asyncio
+
+# Fields the tasks endpoint accepts in the field:operator filter grammar. Deliberately
+# excludes 'status' (task status is not derived by the metadata service, and is being
+# redefined upstream) and tag fields (a task's stored tags are not canonical; the run's
+# tags are grafted on at read time, so SQL tag filtering would match stale values).
+TASK_ALLOWED_FILTERS = [
+    "flow_id",
+    "run_number",
+    "run_id",
+    "step_name",
+    "task_id",
+    "task_name",
+    "user_name",
+    "ts_epoch",
+    "last_heartbeat_ts",
+]
+# Fields whose values must be integer epoch milliseconds.
+_NUMERIC_TASK_FILTERS = {"ts_epoch", "last_heartbeat_ts"}
+
+
+def _split_field_op(key):
+    field, _, operator = key.partition(":")
+    return field, (operator or "eq")
+
+
+def _validate_task_filters(query):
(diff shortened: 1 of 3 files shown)


Aryan95614:
@saikonen this is the tasks-endpoint follow-on to the runs filtering in #481 — same `field:operator` grammar, with status and tag filters intentionally excluded (rationale in the description). Whenever you get a chance to look, much appreciated.

One heads-up: the integration CI job may flag `test_task_attempts_with_attempt_metadata`. That's a pre-existing timing flake in the UI backend (unrelated to this change) — fixed separately on `fix-flaky-attempt-duration-tolerance`.

services/metadata_service/tests/integration_tests/task_test.py
```diff
@@ -287,6 +288,79 @@ async def test_tasks_get(cli, db):
...
+    }
+    assert await _task_ids(cli, old, "?ts_epoch:le=2000") == {
+        old["task_id"],
+        mid["task_id"],
+    }
+    assert await _task_ids(cli, old, "?ts_epoch:ge=2000&ts_epoch:le=2000") == {
+        mid["task_id"]
+    }
+    # user filter narrows to one task
+    assert await _task_ids(cli, old, "?user_name:eq=alice") == {mid["task_id"]}
```
saikonen, line 352: maybe one more test case that covers multiple filters over many fields.

services/metadata_service/api/task.py
```diff
@@ -1,12 +1,64 @@
...
+# Fields the tasks endpoint accepts in the field:operator filter grammar. Deliberately
+# excludes 'status' (task status is not derived by the metadata service, and is being
+# redefined upstream) and tag fields (a task's stored tags are not canonical; the run's
+# tags are grafted on at read time, so SQL tag filtering would match stale values).
+TASK_ALLOWED_FILTERS = [
+    "flow_id",
+    "run_number",
+    "run_id",
+    "step_name",
+    "task_id",
```
saikonen: do we need all the fields in this list? some of these are part of the path, and not the query, so allowing them seems unnecessary

saikonen (approved):

PR #489, opened 2026-08-01
Avoid extra run lookups when returning inherited tags

Aryan95614:
## Summary
- Return inherited run tags for step, task, and artifact reads through an opt-in `runs_v3` join.
- Keep default database-table reads unchanged and enable the join only from metadata API read handlers.
- Cover standard and cursor-paginated read paths with focused unit tests.

#### Test plan
- [x] `python -m pytest services/metadata_service/tests/unit_tests -q` (32 passed)
- [x] Full Postgres 11 test suite (244 passed, 13 skipped)
- [x] Black check for all changed Python files

### DIFF
--- a/services/data/postgres_async_db.py
+++ b/services/data/postgres_async_db.py
@@ -252,6 +252,16 @@ class AsyncPostgresTable(object):
                     conditions=self.trigger_conditions,
                 )
 
+    def _resolve_select_columns(self, with_run_tags: bool = False) -> List[str]:
+        """Columns to select for a single read.
+
+        A method taking the flag rather than a property reading instance state:
+        table objects are shared across concurrent requests, so anything stashed
+        on self can be clobbered by another request across an await. Subclasses
+        that support the run-tags join override this.
+        """
+        return self.select_columns
+
     async def get_records(
         self,
         filter_dict={},
@@ -260,6 +270,7 @@ class AsyncPostgresTable(object):
         limit: int = 0,
         expanded=False,
         cur: aiopg.Cursor = None,
+        with_run_tags: bool = False,
     ) -> DBResponse:
         conditions = []
         values = []
@@ -275,6 +286,7 @@ class AsyncPostgresTable(object):
             limit=limit,
             expanded=expanded,
             cur=cur,
+            with_run_tags=with_run_tags,
         )
         return response
 
@@ -331,6 +343,7 @@ class AsyncPostgresTable(object):
         expanded=False,
         enable_joins=False,
         cur: aiopg.Cursor = None,
+        with_run_tags: bool = False,
     ) -> Tuple[DBResponse, DBPagination]:
         sql_template = """
         SELECT * FROM (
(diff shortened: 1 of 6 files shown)


services/data/postgres_async_db.py
```diff
@@ -905,7 +933,36 @@ async def update_run_tags(
...
+        if getattr(self, "_with_run_tags", False):
+            kwargs["enable_joins"] = True
+        return await super().find_records(*args, **kwargs)
+
+    async def get_records(self, *args, with_run_tags=False, **kwargs):
+        self._with_run_tags = with_run_tags
+        try:
+            return await super().get_records(*args, **kwargs)
+        finally:
+            self._with_run_tags = False
```
saikonen: `get_records` is called by multiple sources on this instance, so there can be a race condition with the `self._with_run_tags`

passing this to the callable directly would be more safe.

services/metadata_service/api/artifact.py
```diff
@@ -7,7 +7,6 @@
     encode_cursor,
     decode_cursor,
 )
-from services.data.tagging_utils import apply_run_tags_to_db_response
```
saikonen, line 10: maybe worth adding a note about why we do not completely remove the helper, instead opting to keep it for the DB writes still.

I assume the main reason is simplicity, compared to having to re-fetch / introduce the join during DB writes?

Aryan95614: Yes, that's the main reason, and there's a second one. I've put both in the `apply_run_tags_to_db_response` docstring so it's answered at the definition rather than at each call site.

The write paths (`add_step` in step.py, `add_task` in task.py) build their response from the INSERT rather than a SELECT, so there's no query to hang the join off. Getting run tags there would mean re-reading the row right after writing it, which is strictly more work than just calling the helper.

The other one is ui_backend_service. `ws.py`, `api/utils.py` and `api/search.py` all still call it on their own read paths, and those don't go through the metadata_service read tables, so moving them over is a separate change and I left it out of scope here.

The note ends by saying the join is the faster path and should be preferred wherever a read already issues a SELECT, so the helper reads as "the case where there is no SELECT" rather than as leftover code.

services/data/postgres_async_db.py
Aryan95614: Agreed. I removed the instance flag in 134d5e3. with_run_tags now travels as a call argument through get_records -> find_records -> _resolve_select_columns, so concurrent reads cannot share or overwrite selection state. get_filtered_tasks_paginated follows the same path, and the focused test interleaves two reads on one table instance and verifies that each builds the expected SQL.

services/metadata_service/api/step.py
```diff
@@ -54,9 +54,8 @@ async def get_steps(self, request):
         """
         flow_id = request.match_info.get("flow_id")
         run_number = request.match_info.get("run_number")
-        db_response = await self._async_table.get_steps(flow_id, run_number)
-        db_response = await apply_run_tags_to_db_response(
```
saikonen, line 58: Just making a note here, as it shouldn't be a blocker. `apply_run_tags_to_db_response` does a run check and returns a 500 error if the run is missing (record orphaned). Most most tables have foreign key constraints to prevent this, but some are missing.

The only effect here is that if an artifact/metadata is for some reason orphaned, and is fetched, then with the left joins it will be returned with NULL tags whereas in the past this would've been an error.

I don't see any major issue with the change here as orphaned records shouldn't be a thing to begin with, but if you come up with a clean fix for the edge-case then feel free to open a follow-up PR :)

saikonen (approved):

PR #490, opened 2026-08-03
Refresh the tag autocomplete cache on run creation events

guts27:
## Summary

The tag autocomplete cache only refreshes on a 5-minute timer, so a new tag can take up to 5 minutes to appear in the dropdown. Each refresh also scans all of `runs_v3`.

This wires the tag cache to `notify.py`, which already receives DB events. As suggested, the event carries the run id and the handler fetches just that run to merge its tags into the cache.

## Changes

- `notify.py`: emit `refresh-run-tags` with `flow_id` / `run_number` on `runs_v3` INSERT
- `ui_server.py`: pass `event_emitter` to `AutoCompleteApi`
- `autocomplete.py`: handler that fetches the single run and merges its tags
- `autocomplete.py`: the periodic refresh now preserves tags added during its scan
- `autocomplete_test.py`: test added

The DB trigger is untouched — `run_number` is already in the notify payload.

## Backward compatibility

- No change to the `/tags/autocomplete` response shape
- No change to the DB trigger
- With `FEATURE_DB_LISTEN_DISABLE=1`, behavior is identical to before (5-minute poll)

## Performance

Local Postgres 11, `runs_v3` with 1.1M rows, 18 distinct tags.

| | Before | After |
|---|---|---|
| New tag visible | 4m 15s | **96ms** |
| Tag removal | up to 5m (periodic) | up to 5m (unchanged) |

The 4m 15s is the average of four measurements (228 / 257 / 266 / 271s) — the poll interval is 5 minutes, so the delay lands anywhere in that range. The 96ms is the average of three (73 / 108 / 107ms), measured from the DB's `clock_timestamp()` at insert to the backend log timestamp when the cache updates.

The two queries involved in refreshing the cache:

| Query | Time | Plan |
|---|---|---|
| `SELECT DISTINCT tag FROM (...)` (periodic) | 2,920ms | Parallel Seq Scan |
| `SELECT tags ... WHERE flow_id=? AND run_number=?` (event) | 3.3ms | Index Scan on `runs_v3_pkey` |

## Design notes

**INSERT only.** `runs_v3` has `last_heartbeat_ts` in `trigger_keys`, so every heartbeat UPDATE fires the same trigger — roughly one every 10 seconds per running run. Listening on UPDATE would mean a lookup per heartbeat even though no tag changed. Tag edits through `metaflow tag add/remove` are handled by the periodic refresh.

**I went without a lock** between the periodic refresh and the handler. I hit the race in a burst test: a periodic refresh that started mid-burst overwrote a tag an event had just added, and the next refresh corrected it.

A lock would have to be held across the full scan (2.9s here, longer as the table grows), and every event handler would wait for it. Instead the periodic refresh snapshots the cache before scanning and re-merges anything added in the meantime:

```python
before = set(self.tags)
res, _ = await self.db.run_table_postgres.get_tags()
if res.response_code == 200:
    added_during_scan = set(self.tags) - before
    self.tags = sorted(set(res.body) | added_during_scan)
```

Both merges run without an `await`, so they can't be interrupted on a single event loop.

**Deletions still rely on the periodic refresh** (up to 5 minutes). To do this as an event, we'd have to check whether the tag is still on any other run, which is the full scan again.

## Verification

- 10 consecutive inserts: all 10 reflected, one handler log each
- Deletions cleared within the poll interval (126s / 183s / 286s)
- `FEATURE_DB_LISTEN_DISABLE=1`: no listener starts, and new tags still appear
- Other autocomplete endpoints and `/runs` unaffected
- Checked in the UI: new tags appear in the dropdown immediately


guts27:
While working on `update_cached_tags` I noticed the cache size log always prints `0 Mb` — `sys.getsizeof` only measures the list itself, not the strings inside it, so the number stays 0 no matter how many tags there are. 
Out of scope for this PR, but happy to open an issue if that's useful.

guts27:
the one integration failure is test_task_attempts_with_attempt_metadata, a pre-existing timing flake unrelated to this change. I'll re-run CI to confirm.

saikonen (commented):
autocomplete/notify changes look good. The diff seems to contain older work as well, can you try to rebase to clean it up?

guts27:
> autocomplete/notify changes look good. The diff seems to contain older work as well, can you try to rebase to clean it up?

the diff and commits now only contain the tag cache changes.
All checks pass.

saikonen (approved):

PR #492, opened 2026-08-06
move cursor fields onto table classes for decode_cursor validation

guts27:
Follow-up to the review comment on #488. Moves the known cursor fields onto the table classes and passes them to `decode_cursor`.

Before, `decode_cursor` only checked for `ts_epoch` and each handler relied on a `KeyError` for the rest. Now each table declares its own `cursor_keys` alongside `primary_keys` / `trigger_keys`, and the handler passes them in.

| Table | cursor_keys |
|---|---|
| flows | ts_epoch, flow_id |
| runs | ts_epoch, run_number |
| tasks | ts_epoch, task_id |
| metadata | ts_epoch, id |
| artifacts | ts_epoch, task_id, name |

These match what each handler was already reading out of the decoded cursor.
The handlers now catch `ValueError` only, since the missing-field case is validated inside the helper.

No behavior change for valid cursors — an incomplete one now returns 400 from `decode_cursor` instead of falling through to a `KeyError`.


services/data/postgres_async_db.py
```diff
@@ -191,6 +191,7 @@ class AsyncPostgresTable(object):
     keys: List[str] = []
     primary_keys: List[str] = None
     trigger_keys: List[str] = None
+    cursor_keys: List[str] = None
```
saikonen: there's a possibility that `decode_cursors` receives a `None` as its cursor_keys if this is not set for a table class as an oversight. Having the default be an empty list would be better in this case.

guts27: Good catch. Thank you. changed the default to an empty list.

saikonen (approved):

PR #494, opened 2026-08-22
Cap task fetch in cache prefetch path

### LINKED ISSUE #473: get_tasks_for_run() fetches all tasks without LIMIT when called from cache prefetch
`get_tasks_for_run()` in `services/ui_backend_service/data/db/tables/task.py` (line 304) calls `find_records()` without passing a `limit`, so it defaults to 0 which means no LIMIT clause in the SQL. With `enable_joins=True` it fires the lateral joins against metadata_v3 and artifact_v3 for every task in the run.

The API layer is fine -- endpoints go through `pagination_query()` which caps at 1000. But this method gets called from the cache prefetch path in `data/cache/store.py` (line 221), which bypasses the API layer entirely.

For a run with a big foreach (hundreds of tasks), this is one unbounded query with lateral joins for each task row. Not sure how often the prefetch actually fires for large runs but it seems like it could be a problem at scale.

roridemonslayer:
Closes #473

`get_tasks_for_run()` called `find_records()` without a `limit`, which omits the `LIMIT` clause entirely (since `0` is falsy in the SQL builder), and combined with `enable_joins=True`, runs an unbounded query with lateral joins against `metadata_v3`/`artifact_v3` for every task in a run.

API endpoints are already protected via `pagination_query()`, which caps at 1000 rows. But the cache prefetch path (`preload_data_for_runs` in `store.py`) calls `get_tasks_for_run()` directly, bypassing that cap entirely — so a run with a large foreach could trigger one unbounded, join-heavy query during prefetch.

## Changes
- Add a `limit` parameter to `get_tasks_for_run()`, threaded through to `find_records()`. Defaults to `0` (no limit) to preserve existing behavior for any other callers.
- Add a `PREFETCH_TASKS_LIMIT` constant in `store.py` (env-overridable, default 1000, matching the API's existing cap) and pass it from the `preload_data_for_runs()` call site.
- Add a regression test (`test_get_tasks_for_run_respects_limit`) covering both the default unbounded behavior and the new limit.

## Test plan
- Added `test_get_tasks_for_run_respects_limit`, verified it fails with `TypeError` against the pre-fix code and passes with the fix, run locally against a real Postgres instance with migrations applied.
- Confirmed no new failures in `tasks_test.py` (4 pre-existing, unrelated heartbeat-timing failures reproduce identically on unmodified `master`).


roridemonslayer:
Hey @saikonen, tagging you since you've touched this file recently and might be a good person to take a look. This PR fixes issue #473, get_tasks_for_run() was missing a limit param so cache prefetch ran an unbounded query with lateral joins for large runs. Added a limit param and a test covering it. Let me know if you'd want anything done differantly.

saikonen (approved):

PR #497, opened 2026-09-02
fix: Artifact pagination edge case

saikonen:
fixes edge case where new paginated route for artifacts could return the latest attempt artifact on a per-artifact basis, not globally for the latest attempt per task. e.g. incorrectly returning
```
artifact_a - attempt 0
artifact_b - attempt 1
artifact_c - attempt 1
```


talsperre (approved):

services/data/postgres_async_db.py
```diff
@@ -1457,6 +1457,92 @@ def _artifact_sql_parts(self, with_run_tags: bool):
...
+    ):
+        # Shared by the run, step and task scoped paginated listings, which differ
+        # only in how they filter.
+        select_sql, values = self._latest_attempt_paginated_sql(
+            filter_dict, cur_ts, cur_task, cur_name, limit, with_run_tags
+        )
+
+        db_response, pagination = await self.execute_sql(
+            select_sql=select_sql, values=values
+        )
```
Aryan95614, line 1534: Not from this PR, but since the three copies are now one helper: if execute_sql fails it returns (error response, None), and the _replace calls below then hit None, so the caller gets an AttributeError turned into a generic 500 instead of the DB error body. The other paginated routes have the same shape, so it's fine as a separate cleanup too, but this is now the one place to guard it.

```suggestion
        db_response, pagination = await self.execute_sql(
            select_sql=select_sql, values=values
        )
        if db_response.response_code != 200:
            # execute_sql returns no pagination on failure.
            return db_response, DBPagination(
                limit=str(limit), offset=0, count=0, page=1
            )
```

Aryan95614 (commented):
Ran the old and new queries side by side on postgres:11 with a synthetic run (4000 tasks, 144k artifact rows over attempts 0 to 2, later attempts dropping some names). The DISTINCT ON version returned 80,000 rows for the run and 10,668 of them came from an attempt older than the task's latest. The window version returns 69,332 rows with none stale and matches a per task max(attempt_id) exactly. The composite index from the July migration is still what the planner picks, and it feeds the WindowAgg without an extra sort.

Not something to change here, but worth knowing: the cursor doesn't reduce the work per page. The latest attempt filter has to see every artifact row of the run before the cursor and LIMIT apply, so page 1 and a deep page both walk all 144k rows (82k shared buffer hits, about 44 ms on this data). The DISTINCT ON version had the same shape (about 27 ms, same buffers), so it's not a regression, just a cost that scales with the run's total artifact rows rather than the page size. If that ever matters, this shape lets Postgres stop at the page boundary and returns the same set. Run scoped first page goes to about 200 buffers and 0.2 ms, deep pages with a cursor to about 400:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS artifact_v3_idx_ts_cursor
    ON artifact_v3 (flow_id, run_number, ts_epoch DESC, task_id DESC, name DESC);

SELECT ... FROM artifact_v3 a
WHERE a.flow_id = %s AND a.run_number = %s
  AND a.attempt_id = (SELECT MAX(b.attempt_id) FROM artifact_v3 b
                      WHERE b.flow_id = a.flow_id AND b.run_number = a.run_number
                        AND b.step_name = a.step_name AND b.task_id = a.task_id)
  AND (a.ts_epoch, a.task_id, a.name) < (%s, %s, %s)
ORDER BY a.ts_epoch DESC, a.task_id DESC, a.name DESC
LIMIT %s
```

The catch is the step scoped listing, which would need its own step prefixed index or it ends up slower than the window version (130 ms on the same data). Happy to put that up as a follow-up if you think it's worth the index.

Two small things. The comment on artifact_v3_idx_latest_attempt in 20260706000000_add_artifact_latest_composite_index.sql still says the index is there for the DISTINCT ON. Maybe reword it so nobody later reads it as dead, something like:

```sql
-- Composite index for the paginated artifact listings. The (flow_id, run_number, task_id)
-- prefix feeds the latest-attempt window (MAX(attempt_id) OVER (PARTITION BY task_id))
-- without a sort. step_name omitted intentionally (breaks by-run).
```

And in the description, "not globally for the latest attempt" reads as one attempt for the whole run, but the reduction is per task, which the test fixture also spells out (the two tasks end on attempts 2 and 1). Worth saying "per task" since the description is what ends up in the release notes.

guts27:
Thank you for catching this! Since this was originally my code, I really appreciate the correction. I was too focused on performance and unfortunately missed the correctness aspect of returning each task's latest-attempt artifacts.

I completely agree that DISTINCT ON (task_id, name) is incorrect. As you pointed out, filtering by artifact name causes an earlier attempt's artifact (like ARTIFACT_C in the test data) to incorrectly survive if a later attempt no longer writes it.

The current approach using the window function with MAX(attempt_id) is the right solution. It also perfectly matches the client-side logic in filter_artifacts_for_latest_attempt. My original tests missed this edge case because every attempt wrote the exact same set of artifacts, so the new test data covers that gap perfectly.

Regarding Aryan's point about the cursor not reducing per-page work: It is true that the DB scan doesn't shrink because the latest-attempt filter must evaluate the entire run before the cursor applies. However, the original goal is still achieved since both the network payload and backend memory footprint are significantly reduced.

